PR #406, opened 2021-03-16
Miner should start with latest waypoint

sm86:
- Miner reads waypoint from key_store.json. 
- Key Store has genesis-waypoint and oper-waypoint. The former is constant, while the later updates with time. 
- The miner should read the oper-waypoint for setting its configs


nelaturuk:
Looks good.

nelaturuk (approved):
Looks good.

liangping (approved):

lpgeiger:
I think this will have to change again in moonshot-cli when that gets merged.

PR #417, opened 2021-03-21
Onboarding gas transfer

lpgeiger:
- [x] user who sends an onboarding transaction will send gas to owner and operator keys of the new validator
- [x] patch tests for balances on both sides.


language/stdlib/modules/0L/LibraAccount.move
```diff
@@ -942,20 +940,17 @@ module LibraAccount {
...
         payee: address
     ) acquires LibraAccount, Balance, AccountOperationsCapability {
-        let bootstrap_amount = 1000000; // 1 gascoin, without scaling representation 
         let payer_addr = Signer::address_of(payer_sig);
         let account_balance = borrow_global_mut<Balance<Token>>(payer_addr);
         let balance_coin = &mut account_balance.coin;
-
-        assert(Libra::value(balance_coin) > bootstrap_amount * 2, Errors::limit_exceeded(EINSUFFICIENT_BALANCE));
+        assert(Libra::value(balance_coin) > BOOTSTRAP_COIN_VALUE, Errors::limit_exceeded(EINSUFFICIENT_BALANCE));
         // Should abort if the
```
nelaturuk, line 951: Can we assert that payee is a validator here?

nelaturuk: Also, the balance_coin should not become 0 after they transfer the value. Can we ensure it is 2x?

language/stdlib/modules/0L/LibraAccount.move
```diff
@@ -942,10 +942,14 @@ module LibraAccount {
         payer_sig: &signer,
         payee: address
     ) acquires LibraAccount, Balance, AccountOperationsCapability {
+        let bootstrap_amount = 1000000; // 1 gascoin, without scaling representation
```
nelaturuk: Cant we have this as a const at the top?

language/stdlib/modules/0L/LibraAccount.move
lpgeiger: > Can we assert that payee is a validator here?
Yes probably, good idea.

> Also, the balance_coin should not become 0 after they transfer the value. Can we ensure it is 2x?
Yes I think there are multiple checks in there for that.

language/stdlib/modules/0L/LibraAccount.move
lpgeiger: Yes good idea

Issue #428, opened 2021-04-02
Rust upgrade to v1.51


simsekgokhan:
Our current Rust version is 1.46. (https://github.com/OLSF/libra/blob/v4/rust-toolchain)

Dev branch:  https://github.com/simsekgokhan/libra/tree/rust_v1.51

## Required actions before the upgrade:
- [ ] Review PR: [rust] upgrade to 1.51 https://github.com/diem/diem/pull/8017
- [ ] Merge from a Rust 1.51 compatible commit of upstream - this will make sure we will not have any build warnings/problems - somewhere close to this commit: https://github.com/diem/diem/commit/e34693b94b06e772c6a07f47da6fb1e34432d764 
- [ ] After merge from upstream, clean build all packages, make sure to there are no build warnings or errors.
- [x] Review Rustc changes between 1.46 to 1.51: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1510-2021-03-25
- [ ] Add `split-debuginfo` into root `Cargo.toml` for ~5x faster incremental debug build times on macOS
```
[profile.dev]
split-debuginfo = "unpacked" # Requires rustc 1.51
```
### 1. ~5x faster the incremental build times with `split-debuginfo`  on macOS
Rust v1.51 brings support for `split-debuginfo` for macOS which enables ~5x faster incremental debug build times.   
In practice from `40s to 7s` for `cargo b -p miner` for a very small comment change.

Config which enabled ~5x faster incremental debug build times on macOS:  
```
root Cargo.toml 
[profile.dev]
split-debuginfo = "unpacked" # Requires rustc 1.51

// ----- These are optional but they will decrease the build time

[profile.dev.package."*"]
# disable debug symbols for all packages except this one
debug = false

.cargo/config
[build] 
rustflags = ["-C", "link-arg=-fuse-ld=zld"]
```
More Details:  
https://github.com/OLSF/libra/wiki/Faster-Build-Times#macos-macbook-pro-mid-2015

### 2. There might be some useful info here:
https://github.com/diem/diem/pull/8017

### 3. Rust 1.51:  
https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1510-2021-03-25

simsekgokhan:
Decided that doing a manual upgrade of Rust version is not a good idea. We should use the version coming from upstream which we currently do.

PR #431, opened 2021-04-05
Rust 1.51 upgrade and enable split-debuginfo feature

simsekgokhan:
This PR is for initial testing, these next actions should be done before we can do the upgrade:  see #428


lpgeiger:
Thanks this is an amazing improvement!

PR #452, opened 2021-04-24
Web monitor enhancements for swarm (db status)

mortonbits:
web-monitor showed wrong status for 'DB boostrapped' if running with a swarm. 

Fixed by adding an explicit property db_path in 'Workspace' config which points to the location where the db is stored in the swarm and per default to the previous location (in the NODE_HOME directory)

previous 0L.toml needs to be deleted and will get recreated properly after:
cargo r -p ol-cli -- --swarm-path=$SWARMTEMPDIR --swarm-persona=alice init


lpgeiger (approved):
Very elegantly done! Looks good to me.

PR #454, opened 2021-04-24
Tx cost params

lpgeiger:
Transactions have different cost profiles. Now applying different cost profiles with Enum. Renaming structs, may cause breaking changes to .0L config files, needs migration, and documentation

- [x] change the tx types `ol/types` 
- [x] all commands in txs must declare the tx type and cost profile
- [x] val-wizard sets cost profile for pre-signed txs
- [x] update account.json fixtures with signed txs


lpgeiger:
closed. included in web-monitor PR

PR #455, opened 2021-04-24
Error module for logging errors

sm86:
Replaced all the assert statements with error module. 
The structure for errors is defined in [wiki](https://github.com/OLSF/libra/wiki/Error-Codes). Please refer to this standard moving forward.


language/move-lang/functional-tests/tests/0L/gas/gas_transfer_check.move
```diff
@@ -17,7 +17,7 @@ fun main(account: &signer) {
 }
 }
 ////////// Transfers should fail ////////
-// check: "Keep(ABORTED { code: 1544,"
```
lpgeiger, line 20: @sm86 Are we able to the error code in these failures?

sm86: Updated, thanks!

language/stdlib/modules/0L/Errors.move
```diff
@@ -16,8 +16,10 @@ address 0x1 {
 /// >TODO: determine what kind of stability guarantees we give about reasons/associated module.
 module Errors {
     /// A function to create an error from from a category and a reason.
-    fun make(category: u8, reason: u64): u64 {
```
lpgeiger, line 19: @sm86 I wonder if you can move the module prefix here, instead of hand coding in each of the modules. I.e. change make() signature to include the module code.

sm86: I considered it, but Move does not support strings, so prefix is not straight forward. So I choose to go with u64 codes

lpgeiger: What I mean is we are using a number code for each module. Why not pass that number here, and concatenate, instead of doing manually?

lpgeiger:
Can you also merge changes from release-v4.3.0?

language/stdlib/modules/0L/Errors.move
sm86: I'm passing the concatenated value - it has module number + function number

PR #459, opened 2021-04-29
added in ol/util a tool to migrate configfiles

mortonbits:
tool to migrate config files

Should be run after version upgrade to avoid manipulating config files manually for already configured nodes. A backup file is created before manipulating the original file.

currently only adds "db_path" into the "[workspace]" section of 0L.toml

potential enhancements:
- update value for an attribute
- remove attribute
- rename attribute

example usage (currently takes no arguments):
cargo run -p ol-util

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -155,6 +155,7 @@ members = [
     "ol-cli",
     "ol/types",
     "ol/fixtures",
+    "ol/util",
 ]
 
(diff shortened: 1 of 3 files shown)


lpgeiger:
This is great @mortonbits. I think we should add a few more fields with names that have migrated. Namely related to TxParams and TxCost fields. Can you take a look at the difference in the files from v4.2.8, and v4.3.0? Very helpful thanks

PR #485, opened 2021-05-07
miner to read configs from swarm_temp and not from HOME/.0L if starte…

mortonbits:
With this update the miner will read configs from the swarm_temp dir and not from $HOME/.0L any longer if started with swarm_path parameter. Also blocks will be stored in swarm_temp directory.

Example usage (no change here compared to before) for alice:
export NODE_ENV="test"
cargo run -p miner -- --swarm-path=.........../swarm_temp --swarm-persona=alice start
 

Prerequisites: "ol-cli init" has to be run before and "blocks" subdirectory has to be pre-filled from fixtures

Example how to establish the prerequisites for alice (after swarm start):
cargo r -p ol-cli -- --swarm-path=.........../swarm_temp --swarm-persona=alice init
mkdir .........../swarm_temp/0/blocks
cp ol/fixtures/blocks/test/alice/* $SWARMDIR/0/blocks

### DIFF
--- a/ol/miner/src/commands.rs
+++ b/ol/miner/src/commands.rs
@@ -14,8 +14,7 @@ use abscissa_core::{
     config::Override, Command, Configurable, FrameworkError, Help, Options, Runnable,
 };
 use std::path::PathBuf;
-use dirs;
-use libra_global_constants::NODE_HOME;
+use crate::entrypoint;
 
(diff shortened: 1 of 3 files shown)


mortonbits:
Cancelling this, because it was based on a wrong branch....

PR #533, opened 2021-06-02
Specify a non-infinite max_open_files for rocksdb

dboreham:


lpgeiger:
Thanks @dboreham will do QA now.

lpgeiger:
@dboreham I just tested your patch. I think the limit may be too low. The node succeeds in starting (with ulimit -n 1000), and fetches a bunch of sync packets, but tries for a very long period of time to parse the blocks. e.g. did not complete one sync job with CPU maxed out for 30mins. 

Let's try 10k?

lpgeiger:
This is patched in upstream. Closing PR

PR #556, opened 2021-06-10
Changing AP module to ignore end_epoch for FIXED_ONCE transaction type

jamesmeijers:
## Motivation

Changing the AP behaviour to ignore end_epoch for FIXED_ONCE transaction type, plus a change to the integration test to fix the issue with intermittent failures in case the test is run without the previously mentioned AP changes. 

yes

## Test Plan

Tested the integration test by running it 10x without seeing failures. Must pass CI tests that will be run.


lpgeiger (approved):
LGTM I'll patch the .json parser for now, so that it always submits 2.

PR #559, opened 2021-06-12
Swarm doc and fixes

mortonbits:
## Motivation

Documentation of swarm was outdated and needed adaption to current usage of ol parameters.
Also added a fix in config.rs and nit_cmd.rs so the source-path is written into 0L.toml.
This was needed for the pilot to run for swarm. Explorer startup also fixed to work form swarm.

## Test Plan

To test the documentation, copy and paste all commands as in ol/documentation/devs/swarm_qa_tools.md

## Related issues:

https://github.com/OLSF/libra/issues/469


mortonbits:
seems like this has already been fixed in main in the recent days. So this PR is not needed any more regarding the code fixes. I will make a new one only for the documentation

Issue #520, opened 2021-05-29
[Mine] Tx panics should not stop miner, instead will queue transaction in backlog.


lpgeiger:
Problem: 
The miner panics when it fails to connect to localhost, when attempting to send TX.

What should happen:
1. The miner should pause and re-attempt connection. 
2. And if fails a second time, continue mining
3. When mining next block is done, resend alll backlog, not just most recent block.

JeffreyQ:
Could addition context be provided around where and when transactions are made? 
Is there a registry outlining all different kinds and types of transaction failures? 
What would be the easiest way to reproduce this json-rpc connection failure?

lpgeiger:
+1 It would be good to document somewhere the different types of tx failures. We don't have this anywhere.

For now a simple disconnection needs to be simulated manually, meaning: kill a libra-node process which was started by swarm and is being used for mining. 

See this: https://github.com/OLSF/libra/wiki/QA-0L-tools-using-Swarm#miner

VenkatTeja:
The miner code has been modified to catch any communication errors with the client during block submission. The miner now maintains the client height and keeps sending backlog after mining new block each time. 

A test has also been written, details of which are in the pull request. (https://github.com/OLSF/libra/pull/566)

lpgeiger:
Awesome! Waiting for CI to run, then I'll QA as well. Thank you!

PR #566, opened 2021-06-16
Fixed: Tx panics should not stop miner (#520)

VenkatTeja:
## Motivation
The is pull request solves the issue raised here: https://github.com/OLSF/libra/issues/520

## Test Plan
A test has been written at https://github.com/VenkatTeja/libra/blob/256ee6ca7aaa0590607790f7f4bfe5f54051d036/ol/miner/tests/integration-submit-tx.rs

The test tries to block the client port using iptables and after sometime checks if miner was able to mine more blocks than before blocking port. Then the port is unblocked and the tests waits for miner to submit the backlog. The test checks again after sometime to check if client is in sync with miner.

### DIFF
--- a/ol/miner/src/block.rs
+++ b/ol/miner/src/block.rs
@@ -120,7 +120,7 @@ pub fn mine_and_submit(
 
             // submits backlog to client
             match backlog::process_backlog(&config, &tx_params, is_operator) {
-                Ok(()) => status_ok!("Success:", "backlog committed to chain"),
+                Ok(()) => status_ok!("Success:", "Proof committed to chain"),
                 Err(e) => {
                     status_warn!("Failed fetching remote state: {}", e);
                 }
@@ -152,7 +152,6 @@ pub fn parse_block_height(blocks_dir: &PathBuf) -> (Option<u64>, Option<PathBuf>
     let mut max_block_path = None;
 
     // iterate through all json files in the directory.
-    println!("{}/block_*.json", blocks_dir.display());
     for entry in glob(&format!("{}/block_*.json", blocks_dir.display()))
         .expect("Failed to read glob pattern")
     {
--- a/ol/miner/src/block.rs
+++ b/ol/miner/src/block.rs
@@ -5,15 +5,14 @@ use crate::{
     delay::*,
     backlog,
     error::{Error, ErrorKind},
-    prelude::*,
-    commit_proof::commit_proof_tx
+    prelude::*
 };
 use byteorder::{LittleEndian, WriteBytesExt};
 use glob::glob;
 use hex::decode;
 use libra_crypto::hash::HashValue;
 use ol_types::block::Block;
-use txs::submit_tx::{TxParams, eval_tx_status};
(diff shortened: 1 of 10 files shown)


ol/miner/src/block.rs
```diff
@@ -104,32 +105,87 @@ pub fn mine_and_submit(
...
-
-            if let Some(ref _node) = config.profile.default_node {
-                match commit_proof_tx(&tx_params, block.preimage, block.proof, is_operator) {
-                    Ok(tx_view) => match eval_tx_status(tx_view) {
-                        Ok(_) => status_ok!("Success:", "Proof committed to chain"),
-                        Err(_) => status_err!("Miner transaction rejected"),
-                    },
-                    Err(err) => status_err!("Miner transaction rejected: {}", err),
+            
+            // submits backlog to client
```
lpgeiger, line 137: We already have a method called backlog::process_backlog(), is there a reason we are not using that here?

ol/miner/src/commit_proof.rs
```diff
@@ -18,10 +18,10 @@ pub fn commit_proof_tx(
     // Create a client object
     let mut client = LibraClient::new(tx_params.url.clone(), tx_params.waypoint).unwrap();
 
-    let chain_id = ChainId::new(client.get_metadata().unwrap().chain_id);
+    let chain_id = ChainId::new(client.get_metadata()?.chain_id);
```
lpgeiger, line 21: I think chain_id is already in TxParams, can you check?

ol/miner/src/block.rs
```diff
@@ -104,32 +105,87 @@ pub fn mine_and_submit(
...
         loop {
+            status_info!(
+                "Latest client block:",
+                format!("{}", remote_height)
+            );
+
             status_info!(format!("Block {}", mining_height), "Mining VDF Proof");
 
-            let block = mine_once(&config)?;
+            let block_result = mine_once(&config);
```
lpgeiger, line 122: If you just call backlog::process_backlog() below, you can simplify this match and just panic.

```
match mine_once(&config) {

                Ok(blk) => block = blk,
                Err(err) => {
                    status_warn!("Mining block {} failed: {}", mining_height, err);
                    continue
                }
            }
```

lpgeiger (changes requested):
Overall looks good, just please check if there is code duplication from the original backlog sending. The backlog always gets sent at the beginning of the miner app starting. So it should just use the same method.

ol/miner/src/block.rs
VenkatTeja: The backlog method might need some modifications though. Will do and push.

VenkatTeja: Closed

ol/miner/src/commit_proof.rs
VenkatTeja: Closed

ol/miner/src/block.rs
VenkatTeja: Closed

lpgeiger (approved):

PR #580, opened 2021-06-24
[WIP] Adding code of conduct.

roamingrabbit:
## Motivation

To add a code of conduct for the  0L community. This template was taken from the upstream libra / Diem repo.

The intent of this  PR is to start a conversation around and agree upon a C.O.C.. The 0L community may decided on a different C.O.C.  Suggestions or ideas around the C.O.C are very much encouraged! please add them here.=)

Yes.

### DIFF
--- a/ol/code_of_conduct.md
+++ b/ol/code_of_conduct.md
@@ -1,0 +1,135 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the
+  overall community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or
+  advances of any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or email
+  address, without their explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Enforcement Responsibilities
+
+Community leaders are responsible for clarifying and enforcing our standards of
+acceptable behavior and will take appropriate and fair corrective action in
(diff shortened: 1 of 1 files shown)


lpgeiger:
At a high level I think this is good. Some more text is needed in order to implement something practical. As per @zmanian, a code of conduct without enforcement is not useful. 

With my OLSF hat on I'll say that the foundation will not be the enforcer of a code of conduct. Also the foundation only publishes software, and doesn't moderate any community tools or endorse any chain that may use that software. There's no police to call.

That said, I think there is a benefit to publishing this document as a template for the different venues of the community, with suggested actions for enforcement.

So each community venue (chat groups, the chain, community funding programs, github) could point to this template as the code of conduct they will adopt. And the enforcement actions available will depend on the tools available and the moderators' abilities (or validators in case of a chain).

Here are my best guesses as to the options available for enforcement per venue.

**Comms channels (discord, twitter, github, etc):**

- Scarlet Letter: public reprimand of an individual. (I personally don't like this, more below)
- Ostracism: Removing the user from community channels.

**A Chain:**

- Confiscation: removing the account from the chain: requires an upgrade needing ⅔ + 1 validators. The venue for enforcing is the chain (Also don't like)
- Status: Removing the voting power (Tower), from the account. 
- Firewalling the user's validator node from consensus. Requires ⅓ + 1 validators.

**Community programs (e.g. Engineering Fund):**

- Voiding any funds due from community programs.


I'll update this list with further suggestions.

lpgeiger:
Separately, regarding my personal preferences:

About #1 "Scarlet Letter" strategies. This option is cheap to do, and as such, easy to get wrong. I think we all have been somehow on both sides of that coin. A community that uses this, has its members live in fear of getting branded. This is a toxic environment to be in, and easily descends into hysteria.  Also, ironically it doesn't solve the problem, and somewhat paradoxically pushes those actors into worse behaviors. I won't be a part of that. 

About #3 "Confiscation" strategies. Any account can be removed by a hard fork, that is, the actors who disagree with a behavior start a new chain, and leave the infractor on a chain of 1 validators. It's feasible, legal, but it's a slippery slope. The "immutability" of blockchains is a fiction; a convention. The data on chain can always be modified by social consensus. But it's a necessary fiction for a foundation of trust. So the convention has become: the state only changes deterministically, and introducing subjectivity to changes leads to chaos. So while this can be done, the routine use of this will leave us with a sad and contentious network.

Also some personal philosophy: it's not clear to me that punishment for the sake of "justice" is actually useful in these communities. What we want is to stop harm from happening, potentially reverse the harm, and prevent it from happening again. So the utility of shaming or confiscating funds is questionable when simply neutralizing an individual, putting a perimeter around them, can be workable. It's how we deal with wild animals. The intent, character, or morality of the alligator is irrelevant, you just want it out of your swimming pool.

mortonbits:
From the code of conduct proposal of this PR I like the following chapters and would be happy if we could put something like that into our repository:

- Our Pledge
- Our Standards
- Scope
- Attribution

Reading the chapters about enforcement responsibilities and enforcement guidelines gives me a bad feeling. Maybe I am too optimistic, but I don’t see the necessity to write these things down here. In case of unacceptable behaviour in the discord channel for example, would’t it be enough to refer to our agreed on code of conduct - if that would ever be necessary at all?

In case of someone behaving "unacceptable" on the chain, it is clear that the normal voting mechanism could always be used to react, if 2/3 of the validators see a need for reaction. Also forking is always possible. I don't see any added value in writing this down. It is clear.

Formulating our code of conduct only in a positive way may also help that things stay positive.

Issue #539, opened 2021-06-03
[Upgrade] Oracle should weight votes based on consensus voting power


0o-de-lally:
- [x] voting power from consensus (tower height) to be used for counting 2/3 majority vote.

jamesmeijers:
#672

Issue #530, opened 2021-06-01
Upstream update 


simsekgokhan:
[Todo]

### Next Actions: 
- Apply config dir patches

#### 1) Missing `genesis` modules 
```
pub mod init;
pub mod node_files;
mod mining;
pub mod seeds;
```
// 0L
https://github.com/simsekgokhan/libra/blob/main/config/management/genesis/src/lib.rs
// Diem
https://github.com/diem/diem/tree/main/config/management/genesis/src

### Notes: 

/// Upstream used: 
diem-core-v1.2.0 - Apr 1, 21
https://github.com/diem/diem/commit/59d951437b8d44bedc9f45534809c2ce0da1fe93

/// New swarm command

```
cargo b -p diem-node -p cli && 
NODE_ENV="test" cargo r -p diem-swarm -- --diem-node target/debug/diem-node -c ~/diem/swarm_temp -n 1 -s --cli-path target/debug/cli
```

simsekgokhan:
### New patches:

1. ol/verifiable_delay/classgroup/src/lib.rs

```
warning: `__gmpz_sizeinbase` redeclared with a different signature
  --> ol/verifiable_delay/classgroup/src/gmp_classgroup/ffi.rs:48:5
   |
48 |     fn __gmpz_sizeinbase(op: &Mpz, base: c_int) -> size_t;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this signature doesn't match the previous declaration
   | 
  ::: ol/verifiable_delay/classgroup/src/gmp/mpz.rs:50:5
   |
50 |     fn __gmpz_sizeinbase(op: mpz_srcptr, base: c_int) -> size_t;
   |     ------------------------------------------------------------ `__gmpz_sizeinbase` previously declared here
   |
   = note: `#[warn(clashing_extern_declarations)]` on by default
```

simsekgokhan:
## Common Errors & Solutions

### 1
```
thread 'functional_testsuite::0L/gas/gas_init.move' panicked at 'Can't fetch account resource for 
0000000000000000000000000B1E55ED', language/testing-infra/e2e-tests/src/executor.rs:198:32
```

Solution: Add 
```
//! account: blessed, 300GAS
```

simsekgokhan:
### New patches:

#### 1. ol/verifiable_delay/classgroup/src/lib.rs 
[...]

Issue #392, opened 2021-03-04
[Upgrade] Upgrade Oracle: move to leader based oracle, and use hashes in votes.


0o-de-lally:
### save space on chain

We would like to use hashes of the stdlib rather than the full stdlib in upgrade votes to save space on chain.

### reduce UI complexity for concurring votes

A voter who is concurring with a proposal need not install source and compile the stdlib to be able to submit a vote.

#### OracleUpgrade.move:

The move module has changes to structs. It must store the binary, and hash for the proposals. But only hash for the votes.

#### Tx Script:
- [x] If a binary is provided: the voting script should decide if this is a a new proposal or if it is concurring.
- [ ] Otherwise a hash must be provided, and will attempt to concur. If no hash is matched it is an error in the script.

## Rust code:
Voters will vote using the TXS app, which can vote with two methods:
- [x] as is currently done, with a binary. And in the case it is a new proposal, the binary must be provided.
- [x] by simply submitting a hash, if they are following the binary.

## Etc
A related Issue: https://github.com/OLSF/libra/issues/537

lpgeiger:
@jamesmeijers Is this straightforward to include in v5? It's somewhat time sensitive.

jamesmeijers:
> @jamesmeijers Is this straightforward to include in v5? It's somewhat time sensitive.

@lpgeiger yeah, pretty straightforward, see PR #672, I'll try to confirm everything early next week.

jamesmeijers:
#672

PR #581, opened 2021-06-25
Vote with hash for upgrades

jamesmeijers:
##

## Motivation

#539 #392 
Updating oracle system to 
1. Only store new stdlibs once, not once per vote
2. Allow validators to vote with hash 
3. Allow validators to delegate their votes (code included and tested, not currently enabled) 
4. Count votes using the weights of each node rather than a one for one vote (code included and tested, not currently enabled)  

Note, depending on the timeline of #530 this PR may need to be pushed back. 

yes

## Test Plan

Wrote new functional tests. Passing previous E2E tests. The integration upgrade test is also passing. I added an update to that test to have one user vote with a hash, however this change has been disabled as it will cause the test to fail when upgrading from v4.3.1 (the test's default) as that version does not have the vote with hash option. 

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/libra/libra/developers.libra.org, and link to your PR here.)


lpgeiger (approved):
Looks great to me. Solid!

PR #588, opened 2021-07-03
Wallet types, Migration framework

lpgeiger:
Introducing setting optional wallet types. These are simple flags which allow users to see the purpose of the wallet.
- [x] `community` and `slow` wallet types created
- [x] slow wallet are the current addresses with transfer limits
- [x] community wallets don't have limits on amount of send and receive.
- [x] community wallets have timed transactions, which validator set can reject.
- [x] autopay module only work on community wallet type
- [x] txs app can send txs to set a wallet to a type
- [x] functional tests in move
- [x] Migration framework
- [x] Migrations stores state of what migrations have run.
- [x] Migrations run on LibraBlock at epoch boundary.


language/stdlib/modules/0L/Migrations.move
```diff
@@ -0,0 +1,86 @@
...
+
+  public fun migrate_community_wallets(vm: &signer) {
+    // find autopay wallets
+    let vec_addr = AutoPay2::get_all_payees();
+    // print(&vec_addr);
+    // tag as 
+    let len = Vector::length<address>(&vec_addr);
+    let i = 0;
+    while (i < len) {
+      let addr = *Vector::borrow<address>(&vec_addr, i);
```
jamesmeijers, line 78: I don't think we should retroactively make wallets community wallets, especially based solely on whether or not they are receiving autopay (I could send an autopay tx to anyone's wallet right now and force them to have a community wallet once this change is done). Just do it as a two part change. First part, people can declare their wallets to be community wallets. Second part, AP doesn't run for non-community wallets. If people don't convert their wallets before the second change, they miss out on payments. If they choose to never convert their wallets, sure they can run away with the money, but at least they won't receive more (and if anyone fails to convert, transaction limits will prevent them from doing anything before we can stop them). 

This has the added advantage that checking whether or not a wallet is a community wallet does not involve searching an unsorted list.

language/stdlib/modules/0L/Wallet.move
```diff
@@ -0,0 +1,411 @@
...
+          is_frozen: false,
+          consecutive_rejections: 0,
+          unfreeze_votes: Vector::empty<address>()
+        })
+      }
+    }
+
+
+    // Utility for vm to set an address as a Community Wallet
+    public fun vm_set_comm(vm: &signer, addr: address) acquires CommunityWallets {
```
jamesmeijers, line 90: As mentioned in Migrations.move, I don't think this function should exist.

language/stdlib/modules/0L/Wallet.move
```diff
@@ -0,0 +1,411 @@
...
+        if (!Vector::contains<address>(&list, &addr)) {
+        
+          let s = borrow_global_mut<CommunityWallets>(0x0);
+          Vector::push_back(&mut s.list, addr);
+        }
+      }
+    }
+
+    // Utility for vm to remove the CommunityWallet tag from an address
+    public fun vm_remove_comm(vm: &signer, addr: address) acquires CommunityWallets {
```
jamesmeijers, line 103: Will this ever be used? I can't think of such a situation unless the whole idea of community wallets is removed.

language/stdlib/modules/0L/Wallet.move
```diff
@@ -0,0 +1,411 @@
...
+  // Note: doesn not need to be a public function, except for use in tests.
+  public fun find(uid: u64, type_of: u8): (Option<TimedTransfer>, u64) acquires CommunityTransfers {
+    let c = borrow_global<CommunityTransfers>(0x0);
+    let list = if (type_of == 0) {
+      &c.proposed
+    } else if (type_of == 1) {
+      &c.approved
+    } else {
+      &c.rejected
+    };
```
jamesmeijers, line 163: Should there be another option to search all three? What if I want to know whether it has been approved or rejected? I won't know which list to search.

language/stdlib/modules/0L/Wallet.move
```diff
@@ -0,0 +1,411 @@
...
+
+    t.veto.count = votes;
+    t.veto.threshold = threshold;
+
+    return votes > threshold
+  }
+
+  // private function to get the total voting power of the validator set, and find the 2/3rds threshold
+  fun calculate_proportional_voting_threshold(): u64 {
```
jamesmeijers, line 264: This is a very high bar to veto a transaction. Maybe that's fine, but under this scheme I doubt any transactions would get vetoed, we have enough trouble getting people to vote for upgrades.

language/stdlib/modules/0L/Wallet.move
```diff
@@ -0,0 +1,411 @@
...
+          Vector::push_back<TimedTransfer>(&mut pending, *t);
+          // TODO: clear the freeze count on community wallet
+          // add to approved list
+        };
+        i = i + 1;
+      };
+      return pending
+    }
+    
+    public fun maybe_reset_rejection_counter(vm: &signer, wallet: address) acquires CommunityFreeze {
```
jamesmeijers, line 300: Why maybe?

jamesmeijers (commented):
I like the community wallet idea, but I have some problems with implementation, especially the vm retroactively making some existing wallets community wallets. I think this should be done by the owners both from a fairness perspective, and also because it makes the program more efficient (no searching a list of community wallets to check if an address is one).

language/stdlib/modules/0L/LibraAccount.move
```diff
@@ -914,8 +915,13 @@ module LibraAccount {
     ): WithdrawCapability acquires LibraAccount {
         //////// 0L //////// Transfers disabled by default
         //////// 0L //////// Transfers of 10 GAS 
-        //////// 0L //////// enabled when validator count is 100. 
+        //////// 0L //////// enabled when epoch is 1000. 
         let sender_addr = Signer::address_of(sender);
+
+        // Community wallets have own transfer mechanism.
+        let community_wallets = Wallet::get_comm_list();
+        assert(!Vector::contains(&community_wallets, &sender_addr), Errors::limit_exceeded(EWITHDRAWAL_EXCEEDS_LIMITS));
```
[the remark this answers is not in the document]
lpgeiger: Yes, I agree it's inefficient. But I don't think it gets used often. I opted to not have a tag on the account because we do need a list of community wallets, for times we need to iterate. Then having an additional struct on an account to tag it as a community wallet seemed excessive. But we can do that.

language/stdlib/modules/0L/Migrations.move
lpgeiger: Yeah totally agree. Take a look, this is not being run actually anywhere. I did this as a test of migrations.

language/stdlib/modules/0L/Wallet.move
lpgeiger: This will need to exist if there is a vote to remove. It's not implemented anywhere, just future proofing.

language/stdlib/modules/0L/Wallet.move
lpgeiger: Good idea.

language/stdlib/modules/0L/Wallet.move
lpgeiger: Agree, I think this parameter is a free variable. @zmanian  has some thoughts on this.

language/stdlib/modules/0L/Wallet.move
lpgeiger: Because there may not be any rejections to reset. Can change name if that's confusing.

language/stdlib/modules/0L/AutoPay.move
```diff
@@ -180,15 +208,20 @@ address 0x1 {
...
-              
-              if (amount != 0 && amount <= account_bal) {
-                if (borrow_global<AccountLimitsEnable>(Signer::address_of(vm)).enabled) {
-                  LibraAccount::vm_make_payment<GAS>(*account_addr, payment.payee, amount, x"", x"", vm);
-                } else {
-                  LibraAccount::vm_make_payment_no_limit<GAS>(*account_addr, payment.payee, amount, x"", x"", vm);
+
+              // check payees are community wallets
+              let list = Wallet::get_comm_list();
+              if (Vector::contains<address>(&list, &payment.payee)) {
```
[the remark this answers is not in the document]
lpgeiger: >Add a function to Wallet module that checks this directly
yeah we have this. will drop it in.

>Check that the wallet is a community wallet when the instruction is added (assuming wallets can't go back and forth)
Yeah I thought I had done that, possibly not. Let me check.

language/stdlib/modules/0L/Wallet.move
```diff
@@ -0,0 +1,323 @@
...
+    public fun get_comm_list(): vector<address> acquires CommunityWallets{
+      if (exists<CommunityWallets>(0x0)) {
+        let s = borrow_global<CommunityWallets>(0x0);
+        return *&s.list
+      } else {
+        return Vector::empty<address>()
+      }
+    }
+
+    public fun is_comm(addr: address): bool acquires CommunityWallets{
```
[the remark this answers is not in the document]
lpgeiger: We need to iterate when we process the transactions of community wallets: LIbraAccount::process_community_wallets

language/stdlib/modules/0L/Wallet.move
lpgeiger: Ok will drop

language/stdlib/modules/0L/LibraAccount.move
```diff
@@ -1038,15 +1044,30 @@ module LibraAccount {
         restore_withdraw_capability(cap);
     }
 
+    //////// 0L ////////
+    public fun process_community_wallets(vm: &signer, epoch: u64) acquires LibraAccount, Balance, AccountOperationsCapability {
```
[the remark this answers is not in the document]
lpgeiger: yep, that's a bug. patched

andrewpdickson:
This sounds great to me!

language/stdlib/modules/0L/Wallet.move
lpgeiger: Will add as a new issue.

earthwindfirewater (approved):

Issue #565, opened 2021-06-16
[Feature Request] Restarting Network from Snapshot


rahulgrover99:
# 🚀 Feature Request
The network may need to re-genesis, based on information from a snapshot of a network.

## Motivation
- For major revisions of the codebase, where VM, stdlib, and framework code are not mutually backward compatible.
- Simulation of onchain state in a local swarm.
- Catastrophic network halt.

## Pitch

The following are the steps that might be followed to implement a solution: 
1. Get state snapshot
2. Get the list of accounts
3. Parse account state for each account on chain
4. Retrieve the state we need, apply transformations, and export to migration.json
5. Create a function such as `vm-genesis/lib.rs encode_genesis_transaction`, which creates genesis transactions from known account states.

lpgeiger:
Thanks for this!

PR #590, opened 2021-07-05
Create genesis from snapshot (#565)

VenkatTeja:
## Motivation
Create genesis from a snapshot. More details are [here](https://github.com/OLSF/libra/issues/565).

# Run
Run the main method from ol/genesis-tools/src/main.rs

## Doubts
1. How to handle accounts whose balance/address was not processed?
2. How to do an real genesis using the blob?

## Test Plan
Create genesis blob and bootstrap db from it. Once db is initiated, check the db for accounts and balances from snapshot.


VenkatTeja:
## Todo
Start swarm using the exact config from the snapshot. This process is running into the following issues:
1. The current test snapshot contains 37 validators and when starting swarm using these, looks like at least one of the node is not starting. Hence, the swarm is not detecting the start of all nodes. 
2. If the above issue is temporarily ignored and we proceed ahead, the swarm continues running. it asks for mnemonic and upon entering, the shell is available. However, not all commands are working. The ```q ar {{ACCOUNT_ADDR}}``` is working but the balance request is failing. This is not a connectivity error but some unexpected info being read.

PR #591, opened 2021-07-05
Query payment events from cli

lpgeiger:
Adds ability of `ol cli` to query events. E.g. to query for sent and received payments.

use with : 

```
// your account
ol query --events-sent
ol query --events-received

// any account
ol --account <account> query --events-sent

// if you don't have all sequence numbers in db, start from a certain sequence number / transaction height
ol query --events-sent --txs-height 100
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5020,6 +5020,7 @@ dependencies = [
  "gumdrop",
  "handlebars",
  "hex 0.4.2",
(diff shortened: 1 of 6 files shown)


soaresa (approved):
Good job! Can't wait to build a wallet view with those queries

andrewpdickson (approved):

PR #601, opened 2021-07-18
TXS timeout and error handling

lpgeiger:
TXS app was timing-out on certain integration tests. The issue was that the time-out was set to 10 secs which is too low. Additionally, many TXS subcommands did not have error handling, and would instead panic, which made it hard to debug.

- [x] longer timeouts in txs::submit_tx
- [x] better error handling in txs app.


soaresa (approved):
Good job!

PR #610, opened 2021-07-30
CLI prevent unreachable sync state height from making the cli panic

lpgeiger:
CLI commands such as `ol start` might crash if an upstream node cannot be reached (e.g. to get a block height reference). This patch handles that error.


soaresa (approved):
Looks good

PR #612, opened 2021-07-31
Refactor tooling libs for use in Tauri GUI

lpgeiger:
For the cross-platform GUI for "light node", some changes were need on ol tooling, so that core functions (e.g. miner::mine_and_submit()) can be called as a library.

Previously the assumption for some functions was that they would only be called from CLI.

- [x] miner: mine_once, mine_and_submit change api so 
- [x] types: config can read configs from a toml file, without abscissa.
- [x] types: config AppCfg::init_app_configs() now can take parameter for statement and ip address, which previously depended on user interaction
- [x] onboard:  keygen command refactor to return mnemonic
- [x] cli: init for swarm refactored to pass args not from commandline.
- [x] remove abscissa prints in functions that are used as library. (pending abscissa upstream fix)


ol/miner/src/commands/start_cmd.rs
```diff
@@ -10,24 +12,24 @@ use reqwest::Url;
...
     )]
     backlog_only: bool,
 
-    // don't process backlog
+    /// don't process backlog
     #[options(short = "s", help = "Skip backlog")]
     skip_backlog: bool,
 
-    // Option to us rpc url to connect
+    /// Option to us rpc url to connect
```
simsekgokhan, line 28: to use?

ol/miner/src/commands/start_cmd.rs
```diff
@@ -10,24 +12,24 @@ use reqwest::Url;
...
     #[options(help = "Connect to upstream node, instead of default (local) node")]
     upstream_url: bool,
 
-    // Option to us rpc url to connect
+    /// Option to us rpc url to connect
```
simsekgokhan, line 32: to use?

ol/miner/src/commands/start_cmd.rs
```diff
@@ -79,7 +81,7 @@ impl Runnable for StartCmd {
             match backlog::process_backlog(&cfg, &tx_params, is_operator) {
                 Ok(()) => status_ok!("Backlog:", "backlog committed to chain"),
                 Err(e) => {
-                    println!("Failed fetching remote state: {}", e);
+                    println!("WARN: Failed fetching remote state: {}", e);
```
simsekgokhan, line 84: I thin this kind of logging with severity is very good and crucial both for dev/production for such a big code base, would be nice to use a logging library where we do not have to write "WARN" manually.

ol/types/src/config.rs
```diff
@@ -114,16 +127,24 @@ impl AppCfg {
...
+        ip: Option<Ipv4Addr>,
     ) -> AppCfg {
         // TODO: Check if configs exist and warn on overwrite.
         let mut default_config = AppCfg::default();
         default_config.profile.auth_key = authkey.to_string();
         default_config.profile.account = account;
 
-        // Get statement which goes into genesis block
-        default_config.profile.statement = what_statement();
+                // Get statement which goes into genesis block
```
simsekgokhan, line 138: Indentation needed?

simsekgokhan (approved):

PR #694, opened 2021-09-20
Move audit 1

simsekgokhan:


language/diem-framework/modules/DiemAccount.move
```diff
@@ -1437,7 +1437,7 @@ module DiemAccount {
         let balance_coin = &mut account_balance.coin;
         // Doubly check balance exists.
         assert(
-            Diem::value(balance_coin) > BOOTSTRAP_COIN_VALUE, // Todo: ">=" ?
+            Diem::value(balance_coin) >= BOOTSTRAP_COIN_VALUE,
             Errors::limit_exceeded(EINSUFFICIENT_BALANCE)
```
simsekgokhan, line 1441: @0o-de-lally Not sure about this one.

simsekgokhan:
@0o-de-lally Any idea about the CI error? Seems like a connection error.

```
...
.Sep 20 08:16:58.931  INFO ureq::unit: sending request POST http://localhost:39501/    
.Timeout waiting for response
```

0o-de-lally:
@simsekgokhan I wonder if you need to pull from main again, since the upgrade integration tests is failing. As you observed it looks like the swarm nodes are failing and cannot be reached by json-rpc. This indicates the upgrade payload is not compatible.

maybe @jamesmeijers  can help on this, since he last looked at the tests.

language/diem-framework/modules/0L/AutoPay.move
```diff
@@ -75,10 +75,9 @@ address 0x1 {
     // It keeps track of all accounts that have autopay enabled and updates the 
     // list as accounts change their Status structs
 
-    // current_epoch value is unused and should be ignored
     struct AccountList has key {
       accounts: vector<address>,
-      current_epoch: u64, //unused
+      // current_epoch: u64, // todo: unused, delete?
```
jamesmeijers, line 80: Ideally yes, delete, but I'm concerned this will cause issues with state migration. I don't know what the migration strategy is though, so maybe its fine.

jamesmeijers:
> maybe @jamesmeijers can help on this, since he last looked at the tests.

Haven't seen that error before...has this been recurring or has it only been tested once? Could be a one-off.

0o-de-lally:
Ok I patched the integration test on main. It was using hard coded values for what branch to test the upgrade against.

language/diem-framework/modules/DiemAccount.move
simsekgokhan: Oh I see. My version is in main now. I will revert to original with a new PR.

PR #700, opened 2021-09-22
Correction

simsekgokhan:


language/diem-framework/modules/DiemAccount.move
```diff
@@ -506,7 +506,7 @@ module DiemAccount {
         assert(MinerState::can_create_val_account(sender_addr), Errors::limit_exceeded(120102));
         // Check there's enough balance for bootstrapping both operator and validator account
         assert(
-            balance<GAS>(sender_addr) >= 2 * BOOTSTRAP_COIN_VALUE, 
+            balance<GAS>(sender_addr) > 2 * BOOTSTRAP_COIN_VALUE,
```
simsekgokhan, line 509: @0o-de-lally 
Does your previous comment about line 1440 "We generally don't want people to have zero balances after an operation. It makes it hard to recover from a problem if a tx needs to be sent and there is no gas." apply here too ? If so, that's my change about.

language/diem-framework/modules/DiemAccount.move
```diff
@@ -1437,7 +1437,7 @@ module DiemAccount {
         let balance_coin = &mut account_balance.coin;
         // Doubly check balance exists.
         assert(
-            Diem::value(balance_coin) >= BOOTSTRAP_COIN_VALUE,
+            Diem::value(balance_coin) > BOOTSTRAP_COIN_VALUE,
             Errors::limit_exceeded(EINSUFFICIENT_BALANCE)
```
simsekgokhan, line 1441: @0o-de-lally Reverting to original. FYI.

language/diem-framework/modules/DiemAccount.move
0o-de-lally: Yes it should apply here as well. Good catch.

PR #706, opened 2021-09-24
Move audit

simsekgokhan:


language/diem-framework/modules/0L/Wallet.move
```diff
@@ -285,6 +303,7 @@ module Wallet {
       return pending
     }
     
+    // Todo: Where is the "maybe" here?
```
simsekgokhan, line 306: @0o-de-lally FYI.

0o-de-lally: good catch. I think that's historic. You can rename.

simsekgokhan: Done.

Issue #711, opened 2021-09-25
[Bug] Audit: privilege escalation possible


0o-de-lally:
Roles:: has an issue with code 0L introduced:
- [ ] new_validator_role_with_proof
- [ ] upgrade_user_to_validator

These functions can be called at any time by the user. So a `user account` can escalate that privilege on their own. Unclear what this affects since a number of other validator configs are necessary to become a validator.

These are typically called by DiemAccount when accounts are created permissionlessly with VDF proofs.

Consider making the setting of the Role, happen at epoch boundary by the VM, upon admission to validator set.

jamesmeijers:
@0o-de-lally #724 fixes this, if I understand your issue with it correctly. The problem is that new_validator_role_with_proof and upgrade_user_to_validator can be called outside of the calls to DiemAccount::upgrade_validator_account_with_proof and DiemAccount::create_validator_account_with_proof, not that the DiemAccount functions can be called by anyone, right? 

I think my change works well, as I understand it, DiemAccount has the special ability to sign as diem root, so now calls to new_validator_role_with_proof and upgrade_user_to_validator require a diem root signature, which can only be provided by the functions in DiemAccount. That said, it's a little odd.

PR #724, opened 2021-09-29
Fix for validator privilege escalation issue. 

jamesmeijers:
Fix for #711


language/diem-framework/modules/Roles.move
```diff
@@ -156,9 +156,11 @@ module Roles {
 
     /// upgrades a user role to validator role
```
simsekgokhan, line 157: Can you also add //////// 0L //////// here?

0o-de-lally (approved):
LGTM

PR #733, opened 2021-10-08
Simple demo spec

simsekgokhan:
Formal Verification demo using Move `spec` syntax


language/diem-framework/modules/0L/Demos.move
```diff
@@ -38,6 +38,14 @@ address 0x1{
           move_to<State>(sender, State{ hist: Vector::empty() });
         }
 
+        // A simple example/demo spec
```
simsekgokhan, line 41: @0o-de-lally Is this what we want? I tried to keep it simple, demo like.

simsekgokhan: Yes, `cargo run --release -p move-prover -- language/diem-framework/modules/0L/Demos.move` checks the specs (this cmd is also in the move prover guide).

PR #761, opened 2021-10-14
clean up dbg print in move error reporting

0o-de-lally:


language/diem-vm/src/errors.rs
```diff
@@ -45,9 +45,6 @@ pub fn convert_prologue_error(
 ) -> Result<(), VMStatus> {
     let status = error.into_vm_status();
```
simsekgokhan, line 47: @0o-de-lally 
I see the https://github.com/OLSF/libra/issues/760 but if we remove this line, we will not be able to see the real error now. I think we should remove this line after/if we solve https://github.com/OLSF/libra/issues/760.

PR #799, opened 2021-11-07
VFN network configurations

0o-de-lally:
- [ ] txs app needs a new subcommand to update on chain discovery info for a validator's nodes. https://github.com/OLSF/libra/issues/904

----- 
The goal of this upgrade is to get all validators on the same node configurations so that mempool, connection attempts, and fullnode network configurations optimize for the network maintaining a full mesh.

The changes are primarily to tooling which generate keys and configurations for the private fullnode used by validators (VFN).

These changes do not affect Move code (no changes to system state or policies). It only upgrades `diem-node`. There is a minor change to diem-node so that it can use the NodeConfig object which now contains the version number which generated the file.

Many (breaking) changes to ol cli - `init`. 
It can now:
- generate the inital 0L.toml with --app.
- create the initial `key_store.json` with --key-store.
- generate a number of node.yaml files: --val, --vfn, --fullnode.
- --seed-peers will fetch the validator fullnodes from chain, and save to file.
- --update-waypoint will fetch from chain to replace base_waypoint in 0L.toml file.

Deprecated:
- `--checkup` and `--fix` are no longer implemented given breaking changes. Perhaps TBD.

Changes to `config/management/genesis/ol_node_files.rs` so that the val, vfn, and fullnode configs can be atomically generated. And refactors throughout to use correct keys and settings. Also ability for node files to include a `seed_peers.yaml` (which can be fetched with `ol init` from chain) to configure files.

Minor change to `onboard kegen` so that `--whoami` will print the different public keys the node depends on.

Global constants are updates, so that there are consistent ports used for fullnode and vfn.

`config/management/genesis/seeds` has changes so that seed peers can be read from file.

Changes to defaults:
mempool: `capacity_per_user` a user can only have 3 txs in the mempool at a given time.
network configs: PING_FAILURES_TOLERATED increased to 10K, MAX_FULLNODE_OUTBOUND_CONNECTIONS incresed from 10 to 100. Hypothetically fixing the "split brain" issues in the December outage.
state-sync: increase the `tick_interval_ms` from 100ms to 1sec.

Patches a bug with `ol` subcommands panicking due to bad unwraps: https://github.com/OLSF/libra/issues/903

FInally: correctly setting the network::upstream settings should allow a validator node, which falls out of the set, to still receive state sync from the vfn nodes, and by extension the public fullnode network.


config/src/config/state_sync_config.rs
```diff
@@ -40,7 +40,7 @@ impl Default for StateSyncConfig {
             mempool_commit_timeout_ms: 5_000,
             multicast_timeout_ms: 30_000,
             sync_request_timeout_ms: 60_000,
-            tick_interval_ms: 100,
+            tick_interval_ms: 1000,
```
0o-de-lally, line 43: @zmanian: proposing to change this

config/src/config/mempool_config.rs
```diff
@@ -24,15 +24,15 @@ pub struct MempoolConfig {
 impl Default for MempoolConfig {
     fn default() -> MempoolConfig {
         MempoolConfig {
-            shared_mempool_tick_interval_ms: 50,
+            shared_mempool_tick_interval_ms: 50, // TODO: Seems too fast.
```
0o-de-lally, line 27: @zmanian: proposing to change this

zmanian (approved):

PR #806, opened 2021-11-12
Onboarding checks if validator is in current set.

0o-de-lally:
Onboarding of validator was missing one check before rate limit. It should have also been checking if an inviting validator is also in the current validator set.

- [x]  DiemAccount checks DiemSystem to see if sender is_validator()
- [x] tests


soaresa (changes requested):
Please solve the conflicts

PR #808, opened 2021-11-12
TowerState check difficulty on initialize

0o-de-lally:
With the new workflow of end-users being able to register without doing a vdf proof, and also being able to initialize TowerState on the submisison of the first proof, there's a bug where the checking of the difficulty was being skipped because of an early return. This code changes the order such that difficulty is checked first on commit_state().


jamesmeijers (approved):
Looks good

soaresa (approved):

PR #831, opened 2021-11-16
Patch description txs transfer

0o-de-lally:
Change help text on txs for sending balance transfers


zmanian (approved):

zmanian (changes requested):
Missing the change

PR #834, opened 2021-11-17
WIP: feat(docker): add updated Docker setup for 0L

nourspace:
## Motivation

I'm building a set of tools to allow easy validator, miners, and full-node onboarding and operating.

## Goals

- Easily create an account and get onboarded
- Easily run tower, full-node, and validator on a VPS that has Docker installed
- Ability to use custom upstream IPs, data directory, etc
- Logging visibility

## Test Plan

Not sure yet how to run tests against the built image. Also how and where to publish it.


docker/0L/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:20.04
```
jackzampolin, line 1: Nit, but would make for a lighter setup: try to go with alpine? IDK if this would be an issue with binaries tho.

nourspace: Good point, thanks.

I'm working on a multi-staged image now. My intuition for using the Easy Mode guide was to be more pragmatic and get something done quick. Now I see that building for source is necessary to correctly run the tower for non-validators.

Let's see whether the built binaries run under alpine. I assume they will have some system deps.

Leibniz137:
Did you try the Dockerfile in `docker/validator/`? Not sure how up-to-date it is

nourspace:
> Did you try the Dockerfile in `docker/validator/`? Not sure how up-to-date it is

@Leibniz137 yes I checked all Docekrfiles and tried to make use of them. They haven't been updated for a long time except for the rebranding from `libra` to `diem`.

chriswessels:
I'd not seen `task` and Taskfiles before. I really like the pattern you've laid out and the ability to spawn interactive workflows (like onboard) that write to the same shared volumes that the node, miner, etc use. It's worth ensuring that the docker images are well decoupled from this `docker-compose` based workflow, so that other orchestration methods (like Helm charts for Kubernetes targets) can easily reuse them. Nice work!

docker/0L/Dockerfile
drawks: If you build them against musl it is possible to get statically linked binaries that will be easily portable. Once you've got static binaries it is possible to have container images that have only the binaries in them and aren't bloated with things like shells and package managers.

nourspace:
> I'd not seen `task` and Taskfiles before. I really like the pattern you've laid out and the ability to spawn interactive workflows (like onboard) that write to the same shared volumes that the node, miner, etc use. It's worth ensuring that the docker images are well decoupled from this `docker-compose` based workflow, so that other orchestration methods (like Helm charts for Kubernetes targets) can easily reuse them. Nice work!

Thanks. `Taskfile` is the hipster version of `Makefile` :D Worth checking out their docs. Pretty simple syntax and has nice features.

I just committed a docker-compose file that shows that separation you mentioned. The result is nice to look at

image

docker/0L/Dockerfile
nourspace: Thank you @drawks for the input. Do you mind giving a hint on how to build against musl. We could add another make subcommand or simply have that in the Dockerfile. I prefer the earlier not to have building commands spread all over the repo.

drawks: @nourspace https://blog.davidvassallo.me/2021/06/10/lessons-learned-building-statically-linked-rust-binaries-openssl/  < this article gives a basic example building a static binary with rust and openssl

nourspace: @drawks great, thank you. I will check it this weekend and see where it gets me.

docker/0L/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    OL_CONFIG="${OL_ROOT_DIR}/.0l.toml"
+
+# Install system prerequisites
+RUN apt-get update && apt-get install -y \
+  curl \
+  && rm -rf /var/lib/apt/lists/*
+
+# Install 0L built binaries [Easy Mode]
+# https://github.com/OLSF/libra/blob/main/ol/documentation/node-ops/validators/validator_onboarding_easy_mode.md
+RUN curl -sL https://raw.githubusercontent.com/OLSF/libra/main/ol/util/install.sh | bash
```
emanuelconunaemme: Why use an external url (with the risk of injections)? Also, might lead to inconsistent CI/CD. E.g.: you modify install.sh and the pipeline won't pick it up here since it's still reading from main.

drawks: I had a similar thought. Ideally the dockerfile would reside in the repo root and would build from the same checkout. I /think/ the current notion is that you could run this Dockerfile standalone and get a build of current master @HEAD

nourspace: Hey @emanuelconunaemme this seems to be an older version where I was trying to use the easy mode still. Please check the latest changes. I'm now using git to clone my own fork. But I'm preparing to start cloning upstream using release tag or simply build directly from it.
I'm yet to conclude where this setup should end up being. The main repo or some other tooling repo that we create.

nourspace:
I'm closing this PR in favour of the work being done here
https://github.com/nourspace/0l-validator-ops/pull/1

I don't want to block anyone from working on the existing docker files while I'm mostly interested in the ops.

PR #857, opened 2021-11-23
update time before validator creation in web monitor

jamesmeijers:
Simple problem, simple fix. Epochs before validators are allowed to create accounts has been increased (const EPOCHS_UNTIL_ACCOUNT_CREATION: u64 = 13; in TowerState.move), but this change has not been reflected in the web monitor.

### DIFF
--- a/ol/cli/web-monitor/src/components/validators/ValidatorModal.svelte
+++ b/ol/cli/web-monitor/src/components/validators/ValidatorModal.svelte
@@ -62,7 +62,7 @@
                 </tr>
                 <tr>
                   <td>can create account</td>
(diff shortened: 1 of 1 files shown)


soaresa:
Thanks @jamesmeijers ! We already have a PR for this issue: https://github.com/OLSF/libra/pull/853

But I do think that we need to change de global constant EPOCHS_UNTIL_ACCOUNT_CREATION to 14 and make the appropriate condition check using ```>=``` instead of ```>``` here: 

image


https://github.com/OLSF/libra/blob/main/language/diem-framework/modules/0L/TowerState.move

I'll be glad to review your changes on that :)

jamesmeijers:
Ah, glad to see it's taken care off. I'll submit a PR for the other thing.

PR #861, opened 2021-11-25
Teams

0o-de-lally:
draft


language/diem-framework/modules/0L/Subsidy.move
```diff
@@ -52,22 +53,98 @@ address 0x1 {
...
+      let i = 0;
+      while (i < Vector::length(members)) {
+        let addr = Vector::borrow(members, i);
+        let one_height = TowerState::tower_for_teams(*addr);
+        print(&100320);
+        print(&one_height);
+        if (one_height > 0) {
+          let payment_to_this_member = FixedPoint32::multiply_u64(
+            value_to_members,
+            FixedPoint32::create_from_rational(one_height, collective_height)
```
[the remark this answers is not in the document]
zmanian: I have a design idea here.

The min tower height is determined by the number of towers who are a members in a team. Basically as you add members to a team the min height increases. I think this would mitigate sybiks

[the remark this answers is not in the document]
0o-de-lally: @zmanian is this the min height of a member to join a team?  Is this like musical chairs in a sense, where people get knocked off a team the  hight to join a team get's progressively higher?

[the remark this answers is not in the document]
zmanian: My mental model is that increasing height requirement on the team can prevent the n+1 team member from join but you don't get kicked out of team if the difficulty becomes higher than your height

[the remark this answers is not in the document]
jamesmeijers: So then even if they make a new team, the rate of growth for that team would be limited…This idea sounds quite promising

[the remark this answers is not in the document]
0o-de-lally: So this means that new members to the network need to go to new teams (teams with fewer members). Does this increase the sybil resistance?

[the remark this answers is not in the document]
zmanian: yes because those new teams will have low aggregate voting power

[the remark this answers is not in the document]
0o-de-lally: So just to be clear. The sybil case we are concerned about is a consensus fault due to 2/3 validators having made sybil collective towers, yes? If so I don't really see that risk being that high (given how the validator tree grows), and the cost to users is quite high, it's hard to explain to end-user miners what the obvious thing to do. And IMO the obvious thing to do is to pick a team captain that runs reliable infrastructure and is participating in the community. Are we ok with that penalty?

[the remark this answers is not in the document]
Vee-Emm: Below is a super crude suggestion.

There is a term called equalizer in APA(pool) leagues .
In the league 5 players whose combined ratings should not exceed 21 form a team  and are pitted against another team with a similar structure. (A player is rated on a scale from 2 to 7 . New players start at 2)


One of the reasons why this works well is because it incentivizes teams to find lower rated players  & as players gain  in ratings they leave and join new teams or form their own ones . But the demand for lower rated (new) players always exist. This organically grows into more and more teams being formed over multiple tournaments .

In 0Ls case , players would be  members with towers , and  rating would be akin to tower heights.

The tradeoff  is that we deviate away from an earlier suggested axiom above that no member should be blocked from joining a team . It can be mitigated if we set the member limit to each team to be slightly higher than  "(available members) / (Total Teams). Teams can negotiate exchange of members too if they so choose . 

Does this  sounds promising ?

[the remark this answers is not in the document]
0o-de-lally: @Vee-Emm This is fascinating. Need to think more about it.
I think the thing to remember is that we want the users to "do the obvious thing", and the obvious thing is to pick the right stewards of the network; join Validators that have an engaged and technical operator.

My proposal is to use @sm86 's idea and do sum of squares to count the collective tower, create an increasing threshold on a per team basis (similar to @zmanian), and increase the rate-limit on account creation (to 4 or 6 weeks)

With these measures I think it's unlikely that one actor can take 1/3rds of consensus seats to halt network, and less likely still to get to 2/3rds and update the state machine.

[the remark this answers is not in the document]
jamesmeijers: I think the sum of squares approach makes sense (or at least sum of some super-linear function, squares is simple, but maybe not optimal, depending on what exactly we want to achieve). Just to confirm, the rate limit would limit the number of teams that can be created, not the number of people that can join teams (which would be unbounded), is that correct?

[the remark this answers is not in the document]
0o-de-lally: The rate-limit I refer to is the creation of new validator accounts, as we have today. So effectively new teams. The increase would be from 2 weeks to 4 or 6 weeks. To give ample time to catch attacks coming.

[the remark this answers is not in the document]
0o-de-lally: Looping back on this discussion. The final implementation uses a threshold of the RMS of the towers. Which achieves the goal of sum of squares per @jamesmeijers

liangping (approved):

language/diem-framework/modules/0L/Subsidy.move
```diff
@@ -49,26 +50,73 @@ address 0x1 {
...
+    }
+
+    public fun split_subsidy_to_team(vm: &signer, members: &vector<address>, value_to_members: u64) {
+      let collective_height = TowerState::collective_tower_height(members);
+      let i = 0;
+      while (i < Vector::length(members)) {
+        let addr = Vector::borrow(members, i);
+        let one_height = TowerState::tower_for_teams(*addr);
+        if (one_height > 0) {
+          let payment_to_this_member = FixedPoint32::multiply_u64(
```
misko9, line 101: If I understand correctly, payment to "active" members will be based off their tower height and not their number of proofs this epoch. Is that right?

language/diem-framework/modules/0L/Subsidy.move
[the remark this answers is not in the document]
misko9: Is the RMS of towers threshold specifically for being able to join a team? 

So phase 1 keeps the validator tower height as the metric for validator set inclusion and consensus. Then, in phase 2 of the rollout, will the collective (not RMS) towers of the validator/team be the metric in order to incentivize team formation?

In phase 2, will the collective delay tower difficulty for teams target 100/max validators to limit subsidy minting?

language/diem-framework/modules/0L/TowerState.move
```diff
@@ -19,9 +19,10 @@ module TowerState {
     use 0x1::ValidatorConfig;
     use 0x1::VDF;
     use 0x1::Vector;
+    // use 0x1::Decimal;
 
-    const EPOCHS_UNTIL_ACCOUNT_CREATION: u64 = 14;
-
+    const EPOCHS_UNTIL_ACCOUNT_CREATION: u64 = 13;
```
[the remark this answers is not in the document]
0o-de-lally: I think this changes is related to use > operator instead of >=

language/diem-framework/modules/0L/Subsidy.move
0o-de-lally: Correct, that's the spec.

0o-de-lally:
bump @zmanian @nelaturuk @jamesmeijers @soaresa @agouin @sm86 @dboreham

language/diem-framework/modules/0L/Teams.move
```diff
@@ -0,0 +1,302 @@
+///////////////////////////////////////////////////////////////////////////
+// 0L Module
+// Teams 
+///////////////////////////////////////////////////////////////////////////
+// Used for coordinating "Teams", equivalent to delegation
+// Teams are groups which engage in work.
+// some of that work is automated such as validation.
+// A Team has a collective "consensus weight", which is used for voting in consensus
```
sm86, line 8: know it might be late but just a thought: should we use consensus weight only for being in top 100 and consider all teams to be equal consensus weight. there is a slight probability of malicious node having higher consensus weight

language/diem-framework/modules/0L/Migrations.move
```diff
@@ -106,4 +106,22 @@ module MigrateTowerCounter {
...
+  use 0x1::Teams;
+  // use 0x1::TowerState;
+  use 0x1::Migrations;
+  const UID: u64 = 2;
+  public fun do_it(vm: &signer) {
+    if (!Migrations::has_run(UID)) {
+      Teams::vm_init(vm);
+      // also initialize relevant state in TowerState
+      // TowerState::init_team_thresholds(vm);
+      Migrations::push(vm, UID, b"MigrateInitTeams");
```
sm86, line 123: CoreAddresses::assert_vm(sender); - if we need this check here

language/diem-framework/modules/0L/Subsidy.move
```diff
@@ -49,26 +50,73 @@ address 0x1 {
...
+      DiemAccount::vm_deposit_with_metadata<GAS>(
+        vm,
+        *captain_address,
+        captain_coins,
+        b"validator subsidy",
+        b""
+      );
+    }
+
+    public fun split_subsidy_to_team(vm: &signer, members: &vector<address>, value_to_members: u64) {
```
sm86, line 94: this is a public function, maybe we need some checks

sm86 (commented):
looks good to me.

language/diem-framework/modules/0L/EpochBoundary.move
```diff
@@ -101,6 +113,29 @@ module EpochBoundary {
         Subsidy::process_fees(vm, &outgoing_compliant_set);
     }
 
+    fun process_burn(vm: &signer) {
```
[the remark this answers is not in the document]
0o-de-lally: I feel like this is written in the rulebook somewhere since it exists since genesis.

language/diem-framework/modules/0L/Teams.move
0o-de-lally: This is interesting. Something for @zmanian

language/diem-framework/modules/0L/Subsidy.move
0o-de-lally: Thanks good catch. Let me take a look.

language/diem-framework/modules/0L/Migrations.move
0o-de-lally: Good catch, let me check.

language/diem-framework/modules/0L/Teams.move
```diff
@@ -0,0 +1,302 @@
...
+    public fun get_team_members(captain: address):vector<address> acquires Team {
+      if (team_is_init(captain)) {
+        let s = borrow_global_mut<Team>(captain);
+        return *&s.members
+      };
+      Vector::empty<address>()
+    }
+
+    // Is this miner above the individual tower height threshold
+    public fun is_member_above_thresh(miner: address):bool acquires AllTeams {
```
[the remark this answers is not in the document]
0o-de-lally: Actually, good catch, we don't have a test for this one, thanks.

language/diem-framework/modules/0L/Subsidy.move
0o-de-lally: fixed with c60c0c213606987d3e537b533af6bd62c1c850d4

language/diem-framework/modules/0L/Migrations.move
0o-de-lally: fixed with 3922dc5afded960d305ab990ba0c1e60959eea6c

language/diem-framework/modules/0L/TowerState.move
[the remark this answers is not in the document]
soaresa: If we maintain this const name it would be better to have the value 14 so when someone creates a query for it the display will make sense without the need to add 1

language/diem-framework/modules/0L/Teams.move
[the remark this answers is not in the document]
0o-de-lally: Fixed with 8b6d06cb2d42bbeecd8f19ae166739baaee57db5

ol/documentation/teams/about_teams.md
```diff
@@ -0,0 +1,95 @@
...
+
+- Opt-in and permissionlesss
+Teams are opt-in, self-assembled groups. People can move freely between teams. What team you are on signals a number of things, but importantly it shows support for a "team captain" and their activities: do they successfully operate nodes, do they participate in network governance, do they provide work and input actively to a network. 
+
+
+- Market dynamics
+The network will not enforce a percentage share between Captains and Members. This is a market, and we'll let the market decide what is appropriate.
+
+- Bias to decentralization
+The network will however bias toward decentralization, and ratchet-up the threshold for a collective tower hight that a validator node is required to have to enter consensus.
```
soaresa, line 25: *height

language/diem-framework/modules/0L/EpochBoundary.move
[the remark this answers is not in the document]
0o-de-lally: I've updated docs a bit: bb203511dc70dce80ff7c220a80de5686de75366

ol/documentation/teams/about_teams.md
0o-de-lally: fixed with bb203511dc70dce80ff7c220a80de5686de75366

language/diem-framework/modules/0L/TowerState.move
```diff
@@ -579,6 +735,41 @@ module TowerState {
...
+    // Permissions: PUBLIC, ANYONE
+    public fun tower_for_teams(node_addr: address): u64 acquires TowerProofHistory {
+      // For a Member's tower height to be counted for a Team for delegation purposes
+      // it must be an active tower (doing work above threshold in an epoch)
+      // and it must not be a new tower, it needs to have a minimum of 7 days equivalent height.
+      if (exists<TowerProofHistory>(node_addr)) { 
+        let s = borrow_global<TowerProofHistory>(node_addr);
+        if (
+          s.count_proofs_in_epoch > Globals::get_epoch_mining_thres_lower()
+          && s.verified_tower_height > TEAM_MEMBER_TOWER_MIN
```
soaresa, line 748: To use operator > here maybe would be better to rename const to TEAM_MEMBER_TOWER_MIN_THRESHOLD

soaresa (approved):
Congrats! Let's start this game

language/diem-framework/modules/0L/TowerState.move
0o-de-lally: fixed in d254dd3a8d46893e90153be0dc3a2e649e665c27

language/diem-framework/modules/0L/Teams.move
```diff
@@ -0,0 +1,302 @@
...
+            member_threshold_epoch: 0,
+            tower_height_rms: 0,
+          }
+        );
+      }
+    }
+
+
+    public fun team_init(sender: &signer, team_name: vector<u8>, operator_pct_reward: u64) {
+
```
jamesmeijers, line 75: Need to confirm that the operator_pct_reward is between 0 and 100

language/diem-framework/modules/0L/Subsidy.move
```diff
@@ -49,26 +50,73 @@ address 0x1 {
...
+      let captain_value = subsidy_granted;
+      if (Teams::team_is_init(*captain_address)) {
+        // split captain reward and send to captain.
+        let captain_pct = Teams::get_operator_reward(*captain_address);
+        // split off the captain value
+        captain_value = FixedPoint32::multiply_u64(
+          subsidy_granted,
+          FixedPoint32::create_from_rational(captain_pct, 100) 
+        );
+        let value_to_members = subsidy_granted - captain_value;
```
jamesmeijers, line 76: I'd confirm that this number is positive and non-zero before continuing, just in case

language/diem-framework/modules/0L/Subsidy.move
```diff
@@ -49,26 +50,73 @@ address 0x1 {
...
         i = i + 1;
       };
     }
 
+    fun check_team_and_pay(vm: &signer, captain_address: &address, subsidy_granted: u64) {
+      // this is a solo validator. Exists during transition to delegation mode. This is a fallback condition to keep the node from halting
+      let captain_value = subsidy_granted;
+      if (Teams::team_is_init(*captain_address)) {
+        // split captain reward and send to captain.
+        let captain_pct = Teams::get_operator_reward(*captain_address);
```
jamesmeijers, line 70: I made another comment in the Teams module, but seeing as this involves the creation of coins, you may want to double check the captains percentage is less than 100 here.

language/diem-framework/modules/0L/Teams.move
```diff
@@ -0,0 +1,302 @@
...
+      let addr = Signer::address_of(sender);
+      // needs to check if this is a slow wallet.
+      // ask user to resubmit if not a slow wallet, so they are explicitly setting it, no surprises, no tears.
+
+      assert(DiemAccount::is_slow(addr), ENOT_SLOW_WALLET);
+        
+
+      // bob wants to switch to a different Team.
+      if (exists<Member>(addr)) {
+        let member_state = borrow_global_mut<Member>(addr);
```
jamesmeijers, line 116: Need to check if the account is in the member list of the old team and if so, remove them

language/diem-framework/modules/0L/Teams.move
0o-de-lally: fixed with: 970536449cf3bc1e2d0b236ffb53fff4297a25c3

language/diem-framework/modules/0L/Subsidy.move
0o-de-lally: fixed with d798090a84aa20dd2f6ae4ee41ccfb337602bfc6

language/diem-framework/modules/0L/Subsidy.move
0o-de-lally: fixed with a90213ae143a5b735f7fc27d8f4f00e2b8f7e4e2

language/diem-framework/modules/0L/Teams.move
```diff
@@ -0,0 +1,326 @@
...
+
+    // a member states his team's preference.
+    // the member only actually joins a team for the purposes of calculating rewards
+    // when they do enough mining in a period
+    public fun join_team(sender: &signer, new_captain: address) acquires Member, Team {
+      let addr = Signer::address_of(sender);
+      // needs to check if this is a slow wallet.
+      // ask user to resubmit if not a slow wallet, so they are explicitly setting it, no surprises, no tears.
+
+      assert(DiemAccount::is_slow(addr), ENOT_SLOW_WALLET);
```
soaresa, line 114: It might be a good idea to wait at least 1 epoch for a new team be able to receive new members because it creates anticipation for Carpe users, who can see new captains donation plans, and avoids validators to create their teams with few members near by the end of an epoch.

language/diem-framework/modules/0L/Subsidy.move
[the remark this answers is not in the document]
0o-de-lally: @misko9 This is covered in the ol/documentaiton/teams. In short yes the Phase 1 does not change qualification or consensus weights. Phase 2 Does. And there are things to work out about the sybil resistance of phase 2, including how the collective difficulty changes.

language/diem-framework/modules/0L/Teams.move
jamesmeijers: I think switch_team would allow a user to be added a team without having first met the mining threshold that is enforced is maybe_activate_member _to_team. Perhaps maybe_switch_team should call maybe_activate_member_to_team in order to prevent this?

sm86 (approved):

language/diem-framework/modules/0L/Teams.move
0o-de-lally: @jamesmeijers please check, I refactored as you suggested.  See: 12aa63f9a1377699f9f77499304271a05d652caf

language/diem-framework/modules/0L/Teams.move
0o-de-lally: @sm86 Can you ticket this as an idea to research in Issues, for Phase 2?

agouin (approved):
Excited for this! LGTM

PR #867, opened 2021-11-27
Update rocksdb crate to support m1 macs

zmanian:
Seems to not break anything and makes builds work on m1 Macs


drawks (approved):
Seems like a sane change

sm86 (approved):

PR #876, opened 2021-12-03
[tools] cli subcommand restore was not updating waypoints in key_store.json correctly

0o-de-lally:
Validators during onboarding were reporting errors when they were already synced to the network and were attempting to  sign blocks. The errors were related to key_store.json not including a recent waypoint. The issue was that when using `ol restore` to fast-forward to an advanced network epoch using a backup, the waypoint in key_store.json was being updated to a previous pattern of namespace.
The fix could have been applied manually, by updating the waypoint field in key_store.json. And then restarting the node in validator mode.

The correction in the code was a one-line fix.

Changes applied to ol/cli/mgmt/restore.rs

- [x] change to the namespace convention. (fixed bug in restore)
- [x] clean up the naming of some fields and functions (rename zip for archive throughout)
- [x] name the downloaded archive correctly as tar.gz
- [x] not update the genesis waypoint, since that is not needed to join consensus successfully

### DIFF
--- a/ol/cli/src/mgmt/restore.rs
+++ b/ol/cli/src/mgmt/restore.rs
@@ -2,7 +2,7 @@
 
 use std::{env, io::Write, process::Stdio};
 use abscissa_core::{status_ok};
-use diem_global_constants::{GENESIS_WAYPOINT, WAYPOINT};
+use diem_global_constants::WAYPOINT;
 use once_cell::sync::Lazy;
 use reqwest;
 use anyhow::{Error, bail, anyhow};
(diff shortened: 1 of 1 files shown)


0o-de-lally:
@JarrydTrokis Kindly take a look a look here ;)

soaresa (approved):
Looks good

JarrydTrokis (approved):
Given my limited understanding of the whole system — this portion of the code looks good to me 👍

Vee-Emm:
Likewise , from what I understand we ensure that the latest waypoint is fetched on a restore instead of the genesis waypoint.

PR #880, opened 2021-12-04
Epochs validating and mining metrics fix

0o-de-lally:
Fix for issue: https://github.com/OLSF/libra/issues/856

The issue was relatively simple, on the epoch boundary the updating of epoch statistics for Tower was happening for all miners, instead of the subset which were validators. This is a reversion, likely due to inclusion of Carpe use cases, and a faulty merge. The tests were also not robust enough for these cases.

The practical matter is that there is a corner case where a miner (not yet a validator) builds a tower for n periods, and then is upgraded to validator (not the workflow we anticipate for validator onboarding) and then instead of epochs_mining_and_validating being 0, it was n.

We take the opportunity to patch this to implement lazy computation of resetting the epoch proof count instead of iterating through the entire list of miners, for a relatively small benefit. THIS WILL REQUIRE CHANGES TO CLIENT, so that users are not confused. More below.

The changes are:

- [x] refactor Tower::reconfig behavior on epoch boundary, we pass the list of compliant validators and iterate over it.
- [x] minor refactor to EpochBoundary.move so we can send the list of compliant validators for the epoch tower statistics.
- [x] correct which list the updating of metrics should occur on (was previously looping the entire miner list, not just validators)
- [x] implement lazy computation for updating non-validator miner epoch count (instead of looping through all miners)
- [x] harden the functional test for Case_3 validators: they should not increment number of epochs_mining_and_validating. There was no test case for this. case_3_reconfig.move
- [x] create a test for the reproduction of the edge case of a miner submitting proofs without being a validator and having the  epochs_mining_and_validating increment. miners_dont_increment_compliant_count.move
- [x] modify JSON_RPC TowerResourceView so that it includes actual_proofs_in_epoch, keeping backwards compatibility.

Web monitor, Explorer, and Carpe will need to update how the count_proofs_in_epoch behaves since moving to lazy computation. The existing JSON_RPC method should be enough,  but a new field should be considered: actual_proofs_in_epoch

If using the JSON_RPC is not possible all  clients will need to check if the latest_epoch_mining is the same as the current epoch, and if so, then the count of proofs is valid. Otherwise the count is stale (the user has not submitted a new proof yet), and the actual proof count in period is 0.


language/diem-framework/modules/0L/TowerState.move
```diff
@@ -503,11 +481,23 @@ module TowerState {
     // returns the number of proofs for a miner in the current epoch
     public fun get_count_in_epoch(miner_addr: address): u64 acquires TowerProofHistory {
```
zmanian, line 504: I feel like you need to store the epoch that get_count_in_epoch was last called in and potentially count over multiple epochs.

You might be quietly allowing state to corrupt if a miner misses an epoch

0o-de-lally: Every time there is an attempt to send a miner tx, we lazily reset the the epoch counter.

https://github.com/OLSF/libra/blob/ratelimit-issue/language/diem-framework/modules/0L/TowerState.move#L250-L256

And it uses the latest_epoch_mining of the user to check if we need a reset:

https://github.com/OLSF/libra/blob/ratelimit-issue/language/diem-framework/modules/0L/TowerState.move#L494-L499

mortonbits:
Looks good to me. 
And the topic "THIS WILL REQUIRE CHANGES TO CLIENT" is no longer needed with the latest implementation of 
**fun get_count_in_epoch**, right? Or do the clients access the variable directly bypassing the getter function?

0o-de-lally:
> Looks good to me. And the topic "THIS WILL REQUIRE CHANGES TO CLIENT" is no longer needed with the latest implementation of **fun get_count_in_epoch**, right? Or do the clients access the variable directly bypassing the getter function?

There's no way that the clients can access the functions in the contracts. They can only access state. So they will need to compute the correct value themselves.

mortonbits (approved):

json-rpc/types/src/views.rs
```diff
@@ -1442,22 +1442,37 @@ pub struct TowerStateResourceView {
...
+    pub actual_count_proofs_in_epoch: u64
 }
 
 impl TryFrom<AccountState> for TowerStateResourceView {
     type Error = Error;
 
     fn try_from(state: AccountState) -> Result<TowerStateResourceView, Error> {
+      let this_epoch = match state.get_configuration_resource()?{
+          Some(cr) => cr.epoch(),
+          None => bail!("cannot get epoch data from account state"),
```
jamesmeijers, line 1457: This is the source of your autopay failure. I believe the issue is that the Account in question does not have a configuration resource. You have to get that resource from the root account.

0o-de-lally: Nice catch!

I think this patched now with my last PR.

PR #886, opened 2021-12-07
Remove operator balance as a check on Audit.move

0o-de-lally:
We have been checking the validators "operator" balance in Audit.move to prevent onboarding an account that was not properly onboarded. This check is no longer necessary, and seems to be doing more harm than good. The validator has the option of running `tower` with owner mnemonic, so it wont be an issue for proof submission.

- [x] change Audit.move to ignore this check
- [x] deprecate functional test for it


jamesmeijers:
> The validator has the option of running `tower` with owner mnemonic, so it wont be an issue for proof submission.

If this is our backup, shouldn't we ensure that the validator has a balance?

0o-de-lally:
> If this is our backup, shouldn't we ensure that the validator has a balance?

This doesn't matter any longer. There's no risk to the network if a validator's operator has no balance. What's happening now is that we have 3-4 prospective validators whose accounts are bricked because of this.

jamesmeijers (approved):
Looks good.

PR #888, opened 2021-12-08
fix to miner underpayment issue

jamesmeijers:
## Motivation

non-validator miners are getting underpaid at epoch boundaries. This is because the global counter of fullnode_proofs, which is being used to calculate the reward for each proof, is not getting reset at the epoch boundary. So, the reward for each proof is calculated as if the sum total of all proofs in the network's history are happening in each epoch, but payouts are only happening based on how many proofs are actually done in the epoch, resulting in severe underpayment. 

Yes

## Related PRs

#880 moves TowerState::epoch_reset to TowerState::test_epoch_reset, this piece of the change should be undone. 

 * This is an urgent change
 * However, we should also consider possible mechanisms to refund miners who have experienced underpayment


zmanian (approved):

zmanian:
I think we should do an upgrade to fix this and punt on retroactive fixes for the time being.

zmanian:
Merging this should block on @0o-de-lally review.

0o-de-lally:
This is close. The issue here is that the global proofs are not the same as the proofs above threshold. So you'll still would be underpaying. Also we need to add some tests. My other PR https://github.com/OLSF/libra/pull/890 addresses the data structure issue. I'm working on this today.

jamesmeijers:
Do we only only pay for proofs above threshold? Or do we pay miners for all their proofs, but only if they mined above a threshold? If it’s the latter I think your change would result in overpayment.

0o-de-lally:
Either one is acceptable. But in principle it's for proofs above threshold. The issue is that without changing the TowerStats the denominator (epoch proofs) includes work done by all miners so the percent would not add up.

zmanian (changes requested):
Block on @0o-de-lally 's alternate PR

PR #890, opened 2021-12-08
[move] change data structure for tracking miner proofs used in Carpe payments

0o-de-lally:
PR resolves the underpayment (incorrect resetting of proofs per epoch) that was affecting Carpe miners receiving exponentially less rewards every new epoch.

These changes include modifications to how TowerState collects proof counts, but also requires a state migration. The state migration needs to happen at the start of the epoch (after the upgrade writeset takes place).

- [x] migrate data structures for tracking miner proofs above threshold TowerStats is now TowerCounter
- [x] TowerCounter includes lifetime_proofs (taken from TowerStats), but also proofs per epoch, and proofs above threshold (which is an issue for correct compensation of Carpe)
- [x] reset TowerCounter correctly after each epoch, functional tests. This was the root cause of the Carpe underpayment bug.
- [x] Block prologue checks for on-the-fly migration on round 3 of every epoch (after an upgrade in round 2)
- [x] tests for mocking and migrating state using Migration.move.


language/diem-framework/modules/0L/FullnodeSubsidy.move
```diff
@@ -17,7 +17,8 @@ address 0x1 {
     use 0x1::TowerState;
 
     public fun get_proof_price(one_val_subsidy: u64): u64 {
-      let global_proofs = TowerState::get_fullnode_proofs();
+
+      let global_proofs = TowerState::get_fullnode_proofs_in_epoch_above_thresh();
```
jamesmeijers, line 21: This function needs divide by 0 protection. Simply need to return 0 if global_proofs is 0

0o-de-lally: done

language/diem-framework/modules/0L/TowerState.move
```diff
@@ -521,6 +572,31 @@ module TowerState {
...
+      state.fullnode_proofs_in_epoch_above_thresh
+    }
+
+    public fun get_lifetime_proof_count(): (u64, u64, u64) acquires TowerCounter{
+      let s = borrow_global<TowerCounter>(CoreAddresses::VM_RESERVED_ADDRESS());
+      (s.lifetime_proofs, s.lifetime_validator_proofs, s.lifetime_fullnode_proofs)
+    }
+
+    public fun danger_migrate_get_lifetime_proof_count(): (u64, u64, u64) acquires TowerStats{
+      if (exists<TowerStats>(CoreAddresses::VM_RESERVED_ADDRESS())) {
```
jamesmeijers, line 591: Should this destroy TowerStats once it is run?

language/diem-framework/modules/0L/FullnodeSubsidy.move
```diff
@@ -17,7 +17,8 @@ address 0x1 {
     use 0x1::TowerState;
 
     public fun get_proof_price(one_val_subsidy: u64): u64 {
-      let global_proofs = TowerState::get_fullnode_proofs();
+
```
jamesmeijers, line 20: I still have a concern. Here we calculate the price we pay per proof as the (total_reward / number_of_proofs_mined_above_threshold), but in payouts, we don't pay miners only for their proofs above threshold, we pay them for all their proofs (assuming they mined over threshold). This will result in the total reward paid out being much higher then intended. In the worst case, all miners mine 7 proofs in the epoch, in which case 7x the intended reward will be paid out. 

We either need to 
- only pay miners for proofs above threshold
- calculate the payout per proof as (total_reward / number_of_proofs_mined), though this will result in underpayment assuming some miner(s) mine below threshold
- calculate the payout per proof as (total_reward / number_of_proofs_mined_by_miners_above_threshold) (this requires a different calculation here)

language/diem-framework/modules/0L/TowerState.move
0o-de-lally: In general I would say yes. But in this case I think it should NOT destroy the old struct. Just so we have it on chain for debugging, since this is the first on-chain migration we are doing. In a subsequent migrations we could do that.

Good catch

language/diem-framework/modules/0L/Mock.move
```diff
@@ -0,0 +1,61 @@
+// Some fixutres are complex and are repeatedly needed
```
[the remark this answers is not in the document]
0o-de-lally: How so? I'm trying to move all of the set up outside of the tests where the code is very busy with this repetitive stuff.

language/diem-framework/modules/0L/FullnodeSubsidy.move
0o-de-lally: So to be clear the way this is implemented is your suggestion #1. We are only paying for proofs subsequent to the threshold being met. I think it's conceptually correct, and doing anything else is tricky. Looking ahead "pay miners for proofs above threshold" also works better for lazy computation, instead of having to loop through all miners an evaluate.

jamesmeijers: Great, the change to EpochBoundary resolves this concern.

language/diem-framework/modules/0L/Mock.move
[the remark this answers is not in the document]
jamesmeijers: Yeah, this is not an issue with what you've done so much as diem...it just doesn't seem right to have this test code be part of production stdlib...I wish there was a compiler flag for test fixtures but move doesn't support this afaik.

jamesmeijers (approved):

PR #894, opened 2021-12-10
[move-vm] secp and keccak hash

coin1111:
## Motivation

1. Add support for ethereum signature verification to move language
        // obtain public key from signature
        // if signature is invalid, then zero key [0u8,20] is returned
        native public fun recover(signature: vector<u8>, message: vector<u8>): vector<u8>;
        // verify signature 
        native public fun verify(signature: vector<u8>, pubkey: vector<u8>, message: vector<u8>): bool;

2. Add support for keccak hash (ethereum sha3)
// Ethereum keccak_256 hash function
        native public fun keccak_256(data: vector<u8>): vector<u8>;
        
Design considerations:
1. For recover/verify
- [x] use older ethers-rs (0.1.3) library, newer ones are not compatible with rust we use 
- [x] recover() returns [0u8,20] (zero key vector) if signature is invalid. This is done to keep return type the same.
- [x] gas for recover/verify is set to 3k as in ethereum erecover op. Both functions are essentially the same (validate signature)
2. For keccak hash
- [x] add support for keccak-256 hashing used compatible with ethereum solidity hash
- [x] add tests 
- [x] use XHash module name to add new hashes in future. This simplifies task of merging with the original diem tree 
- [x] gas for keccak is the same as for sha3_256 (64 units)
- [x] run rustfmt on modified files 
- [x] tests passed : 
cargo test 0L
test result: ok. 155 passed; 0 failed; 248 filtered out
        
yes

## Test Plan

1. Tests for recover/verify are added
2. tests for keccak hash are added


0o-de-lally:
I think this is ready to go. We just need to think a little bit about how this is going to be released. I may keep it off the main branch until we have a new release branch. @dboreham  any thoughts?

0o-de-lally (approved):

PR #898, opened 2021-12-12
Dynamically set the user home directory in systemd unit file

abed-hawa:
## Motivation

Currently users need to update a template manually and start systemd. Goal is to make systemd setup automated.

Related issue: https://github.com/OLSF/libra/issues/543

## Test Plan

I am not familiar with testing systemd or if automated testing is possible, but this was tested on Ubuntu 20.04 (LTS) x64 manually, and it works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -315,10 +315,8 @@ start-full:
 	cargo run -p diem-node -- --config ${DATA_PATH}/fullnode.node.yaml
 
 daemon:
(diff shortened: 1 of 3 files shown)


0o-de-lally (approved):
Very helpful thank you!

drawks (commented):
Would be better with commits squashed

0o-de-lally:
> Would be better with commits squashed

Agree but since we always do squash merges anyways here, so it doesn't tend to bother me.

PR #902, opened 2021-12-15
Prevent panics in stdlib upgrade

vladov3000:
## Motivation
This issue: https://github.com/OLSF/libra/issues/901.
To summarize: Malformed bytes can cause a panic while attempting to upgrade the standard library, causing a network fault.
To fix this, we log any errors and return an error status code instead.

Yes

## Testing

I ran cargo test in the two crates I modified (all tests pass).
And ran an integration test using 
`make -f ol/integration-tests/test-upgrade.mk test`.

### DIFF
--- a/language/move-core/types/src/vm_status.rs
+++ b/language/move-core/types/src/vm_status.rs
@@ -635,6 +635,7 @@ pub enum StatusCode {
     CALL_STACK_OVERFLOW = 4021,
     VM_MAX_TYPE_DEPTH_REACHED = 4024,
     VM_MAX_VALUE_DEPTH_REACHED = 4025,
+    STDLIB_UPGRADE_ERROR = 4026,
 
     // A reserved status to represent an unknown vm status.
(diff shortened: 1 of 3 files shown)


language/move-core/types/src/vm_status.rs
```diff
@@ -635,6 +635,7 @@ pub enum StatusCode {
     CALL_STACK_OVERFLOW = 4021,
     VM_MAX_TYPE_DEPTH_REACHED = 4024,
     VM_MAX_VALUE_DEPTH_REACHED = 4025,
+    STDLIB_UPGRADE_ERROR = 4026,
```
0o-de-lally, line 638: One last change please. The convention for 0L changes is to add a comment above or next to the line, so we know what we have changed for purposes of pulling upstream changes.

```
//////// 0L ////////
```

0o-de-lally:
This is patched in Main now. Thanks for the PR!

PR #905, opened 2021-12-19
[mempool] Ensure mempool resends all txns after disconnect

gregnazario:
Validators don't resend txns after disconnect to ones that have already
confirmed which could cause some issues when many nodes restart or get
stuck.


0o-de-lally (approved):
LGTM, let's try this

PR #911, opened 2021-12-21
[tower] Use diem-logger instead of println

gregnazario:
If diem-logger is used, log levels exist as well as timestamps
which helps the runner know how long it's taking.

### DIFF
--- a/ol/tower/src/backlog.rs
+++ b/ol/tower/src/backlog.rs
@@ -10,6 +10,7 @@ use crate::commit_proof::commit_proof_tx;
 use std::io::BufReader;
 use crate::proof::{parse_block_height, FILENAME};
 use anyhow::{bail, Result, Error};
+use diem_logger::prelude::*;
 
 /// Submit a backlog of blocks that may have been mined while network is offline. 
 /// Likely not more than 1. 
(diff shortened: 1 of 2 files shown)


language/diem-vm/src/diem_transaction_executor.rs
```diff
@@ -490,7 +490,7 @@ impl DiemVM {
 
         let (round, timestamp, previous_vote, proposer) = block_metadata.clone().into_inner();
 
-        println!("========================================== {} ==========================================", round);
+        info!("========================================== {} ==========================================", round);
```
0o-de-lally, line 493: The reason we didn't make this INFO, is because most people run with RUST_LOG=error because of all the noise. So it would suppress this log as well which most people want to see. Any other suggestions how to do it?

ol/tower/src/backlog.rs
```diff
@@ -21,22 +22,22 @@ pub fn process_backlog(
     //let remote_height = remote_state.verified_tower_height;
     let remote_height = get_remote_tower_height(tx_params).unwrap();
 
-    println!("Remote tower height: {}", remote_height);
+    info!("Remote tower height: {}", remote_height);
```
0o-de-lally, line 25: This is good. I like this change.

language/diem-vm/src/diem_transaction_executor.rs
gregnazario: Yeah, then might as well change this back.  Though, it would be nice to include the epoch info which isn't here.

sm86: or we use error!(...) level to log

0o-de-lally (approved):
Thanks for this. Very helpful.

PR #921, opened 2021-12-21
[mempool] Add logging when rejecting a commit

gregnazario:
These logs should help understand when a commit is rejected and removed from the system.

Additionally, helper functions to make it read better (because rejected ones aren't really "committed")

### DIFF
--- a/mempool/src/core_mempool/mempool.rs
+++ b/mempool/src/core_mempool/mempool.rs
@@ -75,9 +75,12 @@ impl Mempool {
             .unwrap_or_default();
 
         if is_rejected {
+            // TODO: Why does the sequence number have to be greater than the current to reject?
             if sequence_number >= current_seq_number {
                 self.transactions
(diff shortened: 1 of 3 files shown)


0o-de-lally:
@gregnazario Looks like the integration tests caught a build error on this file. I tried to patch myself, but don't have all the context. 
https://github.com/OLSF/libra/blob/d7079866f30ed8b8f4aaf5d82131de000e78cc46/mempool/src/shared_mempool/tasks.rs

gregnazario:
> @gregnazario Looks like the integration tests caught a build error on this file. I tried to patch myself, but don't have all the context. https://github.com/OLSF/libra/blob/d7079866f30ed8b8f4aaf5d82131de000e78cc46/mempool/src/shared_mempool/tasks.rs

It should be fixed (i'll let it run through the tests), I just quickly hacked it together and forgot to check if it built correctly.

0o-de-lally (approved):
LGTM

PR #922, opened 2021-12-22
Harden the mempool against a flood of new txs by discarding them earlier in validation

zmanian:
So here is my logic to this change,

https://github.com/OLSF/libra/blob/main/mempool/src/shared_mempool/tasks.rs#L225-L248 rejects old sequence numbers but not new sequence numbers.

but 
https://github.com/OLSF/libra/blob/main/mempool/src/shared_mempool/tasks.rs#L252-L258 but the new sequence numbers get rejected here which is a more expensive validation. 

Rejecting them earlier should help preserve resources

### DIFF
--- a/mempool/src/shared_mempool/tasks.rs
+++ b/mempool/src/shared_mempool/tasks.rs
@@ -223,9 +223,17 @@ where
         .enumerate()
         .filter_map(|(idx, t)| {
             if let Ok(sequence_number) = seq_numbers[idx] {
-                if t.sequence_number() >= sequence_number {
+                if t.sequence_number() == sequence_number {
                     return Some((t, sequence_number));
-                } else {
+                } else if t.sequence_number() > sequence_number{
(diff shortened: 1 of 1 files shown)


zmanian:
One thing about this patch is that I expect because it only blocks txs with newer sequence numbers from other validators. I expect that newer sequence numbers will sit in the mempool of the node that received them until the state has been updated

What this regresses is the ability of a block proposer to propose sequential sequence numbers from a single account.

This more useful for a payments use case than current 0L use cases.

zmanian:
We can revert this when we pull in performance work from upstream

0o-de-lally (approved):
Looks good we did a Canary roll out on this, and seems to mitigate 1003 errors (shared-mem seq number too new). 1002 errors from diem-vm persist. CPU usage is still very high with nodes when carpe fullnode is on.

PR #925, opened 2021-12-25
Raise the log level for peer disconnection from

zmanian:
Raise the log level of peer disconnection events to error. I think this will help diagnose network halts.


0o-de-lally (approved):
LGTM

PR #930, opened 2022-01-01
Tower 72 fix

coin1111:
## Motivation
The issue #1 was found by Barmalei and he suggested the fix as well. While working on #1, I came across issue #2.

There are couple of problems with tower backlog code:
1. tower can submit more than 72 proofs per epoch, however any proofs beyond the first 72 will be rejected with MoveAbort error. Looking in the explorer, this is a fairly frequent error and might contribute to a recent network problems when a bunch of miners accumulated proofs and were unable to send them, when then public node was available submitted all at once and kept sending causing MoveAbort errors thus congesting the network  
2. If a proof cannot be submitted , backlog code selects the second proof in the backlog, thus violating sequence of proofs. This behavior makes no sense, since this proof will never be accepted (since previous has not been sent). This problem was also reported during network incident when proofs were sent out of order , very likely due to this bug.
The fix is to
1. keep track of how many proofs were sent and do not go over 72
2. break backlog if submission error is encountered.
One gotcha found is that count_proofs_in_epoch is not necessarily proofs submitted in the current epoch, thus additional check is required to be able to use this field.

## Test Plan

I tested with a  single miner with a few pre-computed proofs. The code works as expected. 

## Related PRs

 * Justification and breaking nature (who does it affect? validators, full nodes, tooling, operators, AOS, etc.)
 * Comprehensive test results that demonstrate the fix working and not breaking existing workflows.
 * Why we must have it for V1 launch.
 * What workarounds and alternative we have if we do not push the PR.

### DIFF
--- a/ol/tower/src/backlog.rs
+++ b/ol/tower/src/backlog.rs
@@ -62,7 +62,7 @@ pub fn process_backlog(
                 match eval_tx_status(view) {
                     Ok(_) => {}
                     Err(e) => {
-                        warn!("WARN: could not fetch TX status, continuing to next block in backlog after 30 seconds. Message: {:?} ", e);
+                        warn!("WARN: could not fetch TX status. Message: {:?} ", e);
                         break;
                     }
                 };
--- a/ol/tower/src/backlog.rs
+++ b/ol/tower/src/backlog.rs
@@ -53,7 +53,7 @@ pub fn process_backlog(
             while i <= current_block_number.into() && j < remaining_proofs_in_epoch {
                 let path =
                     PathBuf::from(format!("{}/{}_{}.json", blocks_dir.display(), FILENAME, i));
-                info!("submitting proof {}", i);
+                info!("submitting proof {}, in this backlog: {}", i, j);
                 let file = File::open(&path)?;
                 let reader = BufReader::new(file);
                 let block: VDFProof = serde_json::from_reader(reader)?;
--- a/ol/tower/src/backlog.rs
+++ b/ol/tower/src/backlog.rs
@@ -5,11 +5,12 @@ use crate::commit_proof::commit_proof_tx;
 use crate::proof::{parse_block_height, FILENAME};
 use anyhow::{bail, Error, Result};
 use cli::diem_client::DiemClient;
+use diem_client::views::TowerStateResourceView;
 use diem_logger::prelude::*;
 use ol_types::block::VDFProof;
 use ol_types::config::AppCfg;
 use std::io::BufReader;
-use std::{fs::File, path::PathBuf};
+use std::{cmp::max, fs::File, path::PathBuf};
 use txs::epoch::get_epoch;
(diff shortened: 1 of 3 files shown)


ol/tower/src/backlog.rs
```diff
@@ -1,26 +1,43 @@
...
+use ol_types::config::AppCfg;
+use std::io::BufReader;
+use std::{cmp::max, fs::File, path::PathBuf};
+use txs::epoch::get_epoch;
+use txs::submit_tx::{eval_tx_status, TxParams};
 
-/// Submit a backlog of blocks that may have been mined while network is offline. 
-/// Likely not more than 1. 
+// only 72 proofs are allowed per epoch
+const MAX_PROOFS_PER_EPOCH: i64 = 72;
```
agouin, line 18: Can this be fetched from the move globals so that we don't have duplication?

0o-de-lally: My preference is to catch this from the error code which Move throws when the upper bound is reached. `130108`
https://github.com/OLSF/libra/blob/main/language/diem-framework/modules/0L/TowerState.move#L278-L281

More importantly we're missing some error maps of 0L specific errors which the client should identify. See this draft issue I opened. https://github.com/OLSF/libra/issues/933

coin1111: I will fetch it from global store

coin1111: actually after some thinking. I don't think this approach works because with my change tower backlog submission already aborts on any error from submission. So it doesn't matter error code we still should abort submission. This is to solve problem #2 I found with the tower.
And your suggestion to watch for specific error doesn't solve problem #1 in tower - we should not call backlog submission at all if we reached max number of proofs in epoch. Otherwise each tower will keep sending at least 1 proof every 30 mins causing move abort.

0o-de-lally: The thing to keep in mind is that MAX_PROOFS_PER_EPOCH will not always be 72.  And the users need to have the tower continue to make proofs if there's trouble connecting to network or submitting proofs (tower app should not exit).

coin1111: how do I read number 72 from move config? Is there a code snipped to do that? When backup returns (success or failure) , tower proceeds to mine the next proof which  will attempt to be submitted in the next invocation of backlog and so on

0o-de-lally: We have a function in Move which fetches this hardcoded value, but we don't keep it stored in state. So actually we may not be able to get it dynamically through usual APIs

mortonbits: I think it's fine to merge this PR, even if we then will have the allowed proofs per epoch configured at two places. Given the high complexity to make it perfect....

mortonbits (approved):

0o-de-lally:
@coin1111 I've included these changes in https://github.com/OLSF/libra/pull/952 since there were other refactors  in transaction sending. I'll close this PR for now.

PR #938, opened 2022-01-09
[tools] Transaction to update Val configs on chain

0o-de-lally:
A tool was needed to update on-chain discovery (network address) for validators and the validator's VFN fullnode.
1) people will change hosting solutions and need to be able to change IP addresses
2) Validators are migrating to having separate hosts for validator and fullnodes. All the on-chain discovery must be changed to include the new VFN information.

- [x] `txs` subcommand `val-config` which accepts the validator and VFN as arguments. And using mnemonic will submit a tx to change validator configs on-chain.
- [x] cleanup implementation of ValConfigs, so network addresses are consistent.
- [x] Helper to display keys and network info of validator: `ol whoami`
- [x] clean `onboard`and `ol init` commands to exist instead of panic.
- [x] onboarding to ask user to provide IP address of VFN
- [x] wizard gets external IP address for user from https://ifconfig.me


mortonbits:
Looks great. I like to use it in production and then I can also write a doc for "howto migrate to an other IP".

Currently does not compile as FULLNODE_NETWORK_KEY is not found in ol_node_files.rs

nate-staked:
I didn't have trouble with compilation, so might have been fixed by subsequent commits.


This Dockerfile worked fine for me: https://github.com/Leibniz137/0L-node-chart/blob/0o-del-lally-update-val-cfg/docker/Dockerfile

0o-de-lally:
@mortonbits @nate-staked  I've just pushed my final changes. I have both CI and Integration tests passing.

zmanian (approved):

PR #942, opened 2022-01-13
[upgrade]  MoveVM cache issue around upgrades, trigger reconfig

0o-de-lally:
0L's Hot Upgrade mechanism can cause block prologue to abort around the Upgrade Tick (round 2 of each epoch). This does not halt the network, since the block just gets rejected. Any round 2 block with transactions will get rejected. This means that the upgrade may not complete. Previously we tested this by disabling transactions on the main fullnodes for Carpe around the epoch period. The issue is documented here: https://github.com/OLSF/libra/issues/941

The definitive solution thanks to @sblackshear, is to also trigger a reconfiguration so that all the services reset, and among other things clear the Move cache. However there was no way to trigger the reconfiguration without also incrementing an epoch.
This PR addresses that and creates a new type of reconfig: upgrade_recofig(), which is a private function that can be called from rust in the block lifecycle, in this case during the upgrade event. 

Changes:
- [x] Move: add upgrade_reconfig() which only emits a reconfig event, however with the exact same config. Just to signal to the services they need to restart.
- [x] Rust: add a dispatch to Move during the upgrade flow.


zmanian (approved):

zmanian:
This looks good to me and matches me mental model of why the upgrade was failing>

This looks like like it gets deployed by all the nodes upgrading their binaries before the next coordinated upgrade.

PR #944, opened 2022-01-14
[move] autopay bug with percent change instruction type

0o-de-lally:
Patches issues with Autopay calculation when a list is present. Iterating a loop was deducting the denominator in a percentage calculation when it should not have. The solution required a change in state for the module, which in turn required a data migration on all accounts using autopay.

This required creating a function which allows the VM to mock the signature of a User, which has be restricted by the Friend capability.

NOTE to future self: This function needs to be removed as soon as the migration has been successful on chain.


language/diem-framework/modules/0L/Migrations.move
```diff
@@ -106,4 +106,33 @@ module MigrateTowerCounter {
...
+    CoreAddresses::assert_diem_root(vm);
+    if (!Migrations::has_run(UID)) {
+      let enabled_accounts = AutoPay::get_enabled();
+      let i = 0;
+      let len = Vector::length<address>(&enabled_accounts);
+      while (i < len) {
+        let addr = Vector::borrow(&enabled_accounts, i);
+        let account_sig = DiemAccount::scary_wtf_create_signer(vm, *addr);
+        AutoPay::enable_autopay(&account_sig);
+      };
```
sblackshear, line 131: Missing an increment of `i` here I think.

0o-de-lally: Thanks. Sounds like something I would do :)

sm86 (approved):
fast track approve

PR #945, opened 2022-01-14
[upgrade] Upgrade::upgrade_reconfig needs to increment epoch

0o-de-lally:
Manual QA discovered that triggering a reconfig and not incrementing the epoch + 1, would cause the safety-rules to complain on diem-node at the epoch boundary.

we won't be able to keep the same epoch number when we trigger an upgrade_recofig. This means that there will be "phantom" epochs of just 2 rounds whenever there is a successful upgrade. 

The intention is that this be a temporary constraint, and a future upgrade refactor would allow for upgrades to happen during the epoch boundary, and have an ordinary reconfiguration (and thus obviate this implementation).


zmanian (approved):

zmanian:
I assume epoch economics don't run during the phantom epoch

0o-de-lally:
> I assume epoch economics don't run during the phantom epoch

Correct, no accounting or updating of stats.

PR #946, opened 2022-01-15
[core] Diem node upgrade restart

0o-de-lally:
This is Part II of the upgrade Move Cache reset. This includes only Rust changes for v5.0.11, which will be rolled out after the Move chagnes in v5.0.10 have propagated.

Here there are changes for the diem-node, specifically diem-vm, which will trigger a special upgrade_reconfig, which triggers a new epoch event. This will create phantom epochs of 2 rounds. Since the upgrade tick occurs on round 2, and a new epoch event (including incrementing the epoch number) is required to restart services, this epoch will be shorter.
However no accounting or payments occur in this epoch.


mortonbits (approved):

soaresa (approved):
Looks good!

language/diem-vm/src/system_module_names.rs
```diff
@@ -39,13 +39,13 @@ pub static CHECK_UPGRADE: Lazy<Identifier> =
 static UPGRADE_MODULE_NAME: Lazy<Identifier> =
     Lazy::new(|| Identifier::new("Upgrade").unwrap());
 
-// TODO: this will be deprecated in 5.0.11
-pub static RESET_PAYLOAD: Lazy<Identifier> =
-    Lazy::new(|| Identifier::new("reset_payload").unwrap());
+// // TODO: this will be deprecated in 5.0.11
```
dboreham, line 42: Remove comment, or change to : "has been deprecated"?

dboreham (approved):

language/diem-vm/src/system_module_names.rs
0o-de-lally: fixed

PR #978, opened 2022-02-09
Introduce new parameter to omit checking on-chain state for submission of first proof

intuitiveminds:
As discussed in the discord channel, it is currently not possible to submit the genesis proof for a freshly created account purely using the tower app

### DIFF
--- a/ol/documentation/node-ops/tower_mining_VDF_proofs.md
+++ b/ol/documentation/node-ops/tower_mining_VDF_proofs.md
@@ -42,6 +42,14 @@ ol init -u http://<ip-of-a-fullnode>:8080
 onboard user
 ```
 
+To submit the first proof (proof_0.json), you need to start the tower app using a special parameter:
+
+```
+tower start -o -b
+```
+
+After the successful first transfer, continue with starting the tower app (see below).
+
 
 ## Start the tower app
 The tower app will produce VDF proofs or submit proofs, which are created and not yet submitted. The first block to be created is proof_1.json (which takes as the preimage input the sha256 hash of proof_0.json)
--- a/ol/tower/src/backlog.rs
+++ b/ol/tower/src/backlog.rs
@@ -20,10 +20,22 @@ pub fn process_backlog(
     config: &AppCfg,
     tx_params: &TxParams,
     is_operator: bool,
+    omit_remote_check: bool,
 ) -> Result<(), TxError> {
     // Getting remote miner state
     // there may not be any onchain state.
-    let (remote_height, proofs_in_epoch) = get_remote_tower_height(tx_params)?;
+    let mut remote_height = 0u64;
+    let mut proofs_in_epoch = 0u64;
+
+    if !omit_remote_check {
+        let (rem_remote_height, rem_proofs_in_epoch) = get_remote_tower_height(tx_params).unwrap();
+
+        remote_height = rem_remote_height;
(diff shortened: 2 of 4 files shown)


gdoteof:
Commenting since I was following along on discord.

I don't think this is the correct way to solve this.  Right now the client is successfully reaching out the RPC server, which is returning an error, which the client then panics on.

Somewhere in that chain that specific error should not be an error, it should get unwrapped to just explicitly state that the tower doesn't yet exist.

https://github.com/OLSF/libra/blob/e142a7f2a0d483d4fda766882369ced128bf1ee8/ol/tower/src/backlog.rs#L92 

Right here I think there should be another branch of the `match`

that looks like

```rust
    match tower_state {
        Ok(Some(s)) => Ok((s.verified_tower_height, s.actual_count_proofs_in_epoch)),
        Ok(Nothing) => Ok((0, 0)),
        _ => bail!("Info: Received response but no remote state found. Exiting."),
    }
```

I am not a contributor just to be clear; this is just my opinion/perception.

0o-de-lally:
Thanks for this PR. I think there are a lot of good ideas here. The solution here to your problem is to make sure you have an RPC upstream to submit to. I have a related PR for fetching a seed list of upstreams: https://github.com/OLSF/libra/pull/980

That all said. This is really great Rust code! I'd like to repurpose this so that the user can submit proof #X from the command line. Which I think is generally useful. Are you up for trying that?

gdoteof:
> Thanks for this PR. I think there are a lot of good ideas here. The solution here to your problem is to make sure you have an RPC upstream to submit to. I have a related PR for fetching a seed list of upstreams: #980

I am pretty certain this is not true @0o-de-lally (I originally said the same to @intuitiveminds ). 

If there is no tower state the upstream server itself returns: `{ code: -32000, message: "Server error: could not get tower state", data: None }`; the behavior is the same if there are no upstream nodes to connect to because the specific error is swallowed; however this behavior also happens in the case where the upstream node simply fails to find the tower state (because there is no tower yet).

intuitiveminds:
> Thanks for this PR. I think there are a lot of good ideas here. The solution here to your problem is to make sure you have an RPC upstream to submit to. I have a related PR for fetching a seed list of upstreams: #980
> 

Both you and @gdoteof are more versed in this code, given that I just started digging in here, but I think the RPC upstream might not be the problem here as @gdoteof pointed out. It's more the way this particular scenario is handled.

> That all said. This is really great Rust code! I'd like to repurpose this so that the user can submit proof #X from the command line. Which I think is generally useful. Are you up for trying that?

Sure, I can do that. This assumes that our users are a bit versed (e.g. they at least know where to find the proofs in the filesystem etc.), otherwise we would have to provide a way to list the local proofs, show the remote tower height etc. (similar to Carpe, but just in the terminal). Do you only want this feature to submit a specific proof or all proofs starting at a specific one?

0o-de-lally:
Thanks I think this will be very helpful. You'll see that the files are always stored in the same place, and those settings are in the app config file (0L.toml). Looking forward :)

0o-de-lally:
@intuitiveminds Hi there, been a while, since thie was opened.

Is this change still valid? If so I can review it again. BTW can you also check your other PR and let me know if it's also relevant and ready for review?

intuitiveminds:
This PR is pretty much superseded by https://github.com/OLSF/libra/pull/1001, so closing it.

PR #980, opened 2022-02-09
[tools] rpc-playlist initialization with `ol init`

0o-de-lally:
- [x] add rpc-playlist option to `on init` so that users can easily update the `upstream-peers` from command line.
- [x] remove unwraps in `ol start` which are causing panics


0o-de-lally:
bad merge

PR #998, opened 2022-02-22
[move] Increase the proof-of-burn and make dynamic

0o-de-lally:
We knew it would happen: Prisoner's dilemma defection. Someone decided to not donate to community funds and they begin to get depleted. So stronger consensus is needed to fund the public good wallets.

This upgrade is very simple. Using existing infrastructure we just change one parameter.

Since Genesis there has been a 1 coin Epoch Burn for Validators. Validators can opt for this Burn to go to an index of community wallets (set algorithmically based on prior contributions).

The upgrade changes the  Epoch Burn, amount from a flat 1 GAS coin to a variable amount. This proposal sets it to 25%.

The expected effect of this is that community wallets can be more meaningfully funded from Epoch Burns. Note: Burning to the "Router" of community wallets is optional, otherwise the default is a normal burn (destroy the coins). Validators may be encouraged to donate to certain wallets so as to rebalance the index. More details are written in the Match Burn documentation.

- [x] increase value to 50% of previous epoch's value
- [x] expose transaction script for changing preferences
- [x] add txs subcommand to set burn preference.

https://github.com/0o-de-lally/libra/blob/scrooge/ol/documentation/accounts/match_burn.md


mortonbits:
Nice idea about the router and the opt-in between token burn or sending "autopay" to the router. But for me looks like some functionality (especially tooling) is missing, like how validators can opt-in to the different variants. If this change will go live, all validators with currently 50% autopay need to adapt their config, otherwise there will be 75% "tax" for them. But maybe this is intended to punish validators which do not follow what's going on.

Maybe a simpler implementation of how to achieve 50% "tax" including the possibility for token burn while keeping the current tooling and conecpt could be to implement a special "burn" address. So this address can be used in the existing autopay mechanism as target. On entering validator set and before subsidy distribution we should just check for autopay being set to 50% or more.

0o-de-lally:
So there are a couple tasks left, which are low hanging:

- [ ] increase value to 50% of previous epoch's value
- [ ] expose transaction script for changing preferences
- [ ] add txs subcommand to set burn preference.
- [ ] set a timer/epoch for when it goes into effect.

sm86 (approved):
fast track approve

PR #1052, opened 2022-03-30
Update validator set selection

jamesmeijers:
## Motivation

Currently, 100 potential validators are selected, and ineligible ones are removed from that list. Changing the code to select 100 eligible validators with the most weight, not just the eligible validators from the 100 with the most weight. 

Yes

### DIFF
--- a/language/move-lang/functional-tests/tests/0L/reconfiguration/reconfig_case_2_rejoin.move
+++ b/language/move-lang/functional-tests/tests/0L/reconfiguration/reconfig_case_2_rejoin.move
@@ -227,7 +227,7 @@ script {
         Vector::push_back<address>(&mut voters, @{{bob}});
         Vector::push_back<address>(&mut voters, @{{carol}});
         Vector::push_back<address>(&mut voters, @{{dave}});
-        // Vector::push_back<address>(&mut voters, @{{eve}});
+        Vector::push_back<address>(&mut voters, @{{eve}});
         Vector::push_back<address>(&mut voters, @{{frank}});
 
         let i = 1;
--- a/Makefile
+++ b/Makefile
@@ -370,13 +370,13 @@ ifdef TEST
 	@echo NAMESPACE: ${NS}
 	@echo GENESIS: ${V}
 	@if test ! -d ${DATA_PATH}; then \
-		echo Creating Directories \
+		echo Creating Directories; \
 		mkdir ${DATA_PATH}; \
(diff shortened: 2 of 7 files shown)


language/move-lang/functional-tests/tests/0L/reconfiguration/reconfig_case_2_rejoin.move
```diff
@@ -227,7 +227,7 @@ script {
         Vector::push_back<address>(&mut voters, @{{bob}});
         Vector::push_back<address>(&mut voters, @{{carol}});
         Vector::push_back<address>(&mut voters, @{{dave}});
-        // Vector::push_back<address>(&mut voters, @{{eve}});
```
jamesmeijers, line 230: @0o-de-lally when I made the change, this test started failing the assert in line 252, eve was hitting case 4 rather than case 2. Case 2 means eve was validating but not mining, but with this line commented out, Eve was not validating, so case 4 would be correct. Am I missing something here?

0o-de-lally: @jamesmeijers  Yes this test needs to be cleaned up. The text comments refer to EVE as the Case 2. But we're actually testing FRANK. So your change is correct (but I'm not sure why it would have been passing before your change).

The check on line 242 is not necessary, since at that point in the test all validators will be case 2 (since we don't mock the mining yet).

I also noticed that the two scripts in 94 and 116 are duplicates. I think it's harmless though.

jamesmeijers: Glad to hear my understanding is right, very odd/concerning that is was passing before though.

0o-de-lally:
@jamesmeijers @sm86  I have a similar change in https://github.com/OLSF/libra/pull/1071, but chooses validators with highest proposals in given epoch, rather than tower height.

Can you check if this implementation satisfies the spec.

https://github.com/OLSF/libra/blob/17a28c64c93d2c4ea510be628588f7f84aaf5fc6/language/diem-framework/modules/0L/EpochBoundary.move#L202

https://github.com/OLSF/libra/blob/17a28c64c93d2c4ea510be628588f7f84aaf5fc6/language/diem-framework/modules/0L/Stats.move#L250

0o-de-lally:
This was merged with https://github.com/OLSF/libra/pull/1071

PR #1059, opened 2022-04-05
Make whole carpe

jamesmeijers:
Move stdlib updates for make whole. 

Alternative to #1040 which automates the payments. This version requires miners to claim their rewards.


0o-de-lally:
@jamesmeijers  Good work. I think this is the correct implementation. I'm sure @soaresa can quickly get a transaction that works in Carpe for this.

0o-de-lally (approved):
LGTM

PR #1061, opened 2022-04-10
Update start_a_full_node.md

ashiiix:
Just a doc update

### DIFF
--- a/ol/documentation/node-ops/auto_restart_validators.md
+++ b/ol/documentation/node-ops/auto_restart_validators.md
@@ -1,0 +1,35 @@
+# Automatic restart of Validators
+
+## Assumptions
+
+- The command `killall` is installed
+- The ol binaries are in ~/bin
+
+## Instructions to start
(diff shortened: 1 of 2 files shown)


mortonbits (commented):
Thanks, make sense to document port 6178 needs to be accessible

codilion:
I would suggest to remove the offline nodes too. 
Additionally, as the last step, change address for `json_rpc` in the `fullnode.node.yaml` to
 `address: 0.0.0.0:8080`

ashiiix:
Yes I completly forgot that point, my bad.

mortonbits:
> As the last step, we have to add the change address for `json_rpc` in the `fullnode.node.yaml` to `address: 0.0.0.0:8080`

Not sure if this is a good idea - in the past we had problems with this setup. Maybe @0o-de-lally can give some hint about this.

0o-de-lally (approved):

Issue #1063, opened 2022-04-11
BUG: Validator failover not implemented to spec


sm86:
Investigate the validator set when everyone is inactive. Currently it goes to 100 nodes. 

Ideally should have the current validator set if no one qualifies

jamesmeijers:
PR for this #1069

sm86:
this might not address the issue we have. 
If network is halted and none of the validators submit proofs, we continue to have all top_accounts

sm86:
### Proposed algo: 
proposed set  = the list of not jailed eligible validators (submitted threshold proofs)
if size(proposed set) < 3:
    top 50% of previous epoch's validator set sorted by tower height

// the above is recursive, gradually reducing the size of network while halted. 
base case: ensure at least 10 validators in the set

### Assumptions
Assumes at least 7 out of top 10 validators (sorted by tower height) are active

0o-de-lally:
Thanks @sm86  this is useful. Can you propose some sample Move code?

0o-de-lally:
@jamesmeijers Thanks for the fast turn around on this! What do you think of the feasibility of Sha's proposal?

jamesmeijers:
> If network is halted and none of the validators submit proofs, we continue to have all top_accounts
I don't think so, in this case it will revert to the old validator set. @0o-de-lally unless I'm wrong that `ValidatorUniverse::get_eligible_validators(vm);` returns the full list of potential validators regardless of their mining in the last epoch. The reason I added the loop to check if validators are in that set is in case a validator removed themselves from the set. 

That said, I think the gradual reduction proposal seems reasonable.

sm86:
might be a special case but right now we have 100 validators and none of them submit proofs since network halted - so we might end up with same validator set every epoch and never reach consensus

sm86:
@0o-de-lally and @jamesmeijers let me share some Move code soon

0o-de-lally:
Fixed with https://github.com/OLSF/libra/pull/1118

PR #1069, opened 2022-04-15
update epoch boundary to failback to current validator set

jamesmeijers:
## Motivation

#1063

### DIFF
--- a/language/diem-framework/modules/0L/EpochBoundary.move
+++ b/language/diem-framework/modules/0L/EpochBoundary.move
@@ -24,6 +24,7 @@ module EpochBoundary {
     use 0x1::DiemAccount;
     use 0x1::Burn;
     use 0x1::FullnodeSubsidy;
+    use 0x1::ValidatorUniverse;
 
     // This function is called by block-prologue once after n blocks.
     // Function code: 01. Prefix: 180001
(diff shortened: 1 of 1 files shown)


drawks (approved):

language/diem-framework/modules/0L/EpochBoundary.move
```diff
@@ -146,13 +147,35 @@ module EpochBoundary {
         };
 
         // If the cardinality of validator_set in the next epoch is less than 4, 
-        // we keep the same validator set. 
-        if (Vector::length<address>(&proposed_set) <= 3) proposed_set = *&top_accounts;
+        // we keep the same validator set, removing validators who have removed themselves 
+        if (Vector::length<address>(&proposed_set) <= 3) {
+            let eligible_validators = ValidatorUniverse::get_eligible_validators(vm);
+            proposed_set = DiemSystem::get_val_set_addr();
```
roamingrabbit, line 153: Could you elaborate in the comment the detailed  intent? It's not entirely clear to me when for example a validator would be in the `eligible_validators` (validator that submitted a proof in the last epoch) but not in the `current_validator_set` (all the validators from the last epoch).  I'm probably being dense, but a bit more explanation would be helpful!  e.g.
"if we do not have enough validators in the proposed_set, recalculate the proposed_set, ignoring the `top_accounts` and jailed accounts, and take all the current validator set for the epoch `DiemSystem::get_val_set_addr()` and the validators who "submitted a mined proof correctly in the current epoch".

0o-de-lally:
@jamesmeijers @roamingrabbit I've incorporated these changes in my other work on hardening the epoch transition here: https://github.com/OLSF/libra/pull/1071. So will close this PR for now.

PR #1070, opened 2022-04-15
Reconfiguring validator set 

sm86:
#1063 #1069 

this is algorithm but incomplete


0o-de-lally:
@sm86 I see you are implementing this as a special case. But I think the limit to seat rotation is just as important during steady state. I.e. we don't need to create controls to turn it on and off during "rescue operations" only.

PR #1071, opened 2022-04-15
[tools] Network Recovery Tools and Epoch Boundary Hardening

0o-de-lally:
The ability to apply manual transactions to a database at rest was untested. Many of the types of manual transactions we needed were missing: Update the stdlib, and bulk update validators.

After creating those types of TXS, there's an issue with how writsets are applied to DBs at rest.

Backups from state-snapshot have not been compatible with db-bootstrapper. This is because epoch reconfigurations (which are necessary on every db-bootstrapper command) were not being triggered because the stdlib deduplicates multiple reconfigurations by checking that TimestampNow =! Last reconfiguration time. However in snapshots at an epoch boundary those will be the same, and you'll be prevented from applying any writesets to an epoch snapshot.

changees to writeset-generator:
- [x]  new writeset tool to bulk update validators
- [x]  new writeset for a fresh stdlib
- [x]  writeset to trigger upgrade reconfig
- [x]  new writeset for a fresh stdlib
- [x] writeset to increment timestamp by 1ms to get around reconfig deduplication bug.


language/diem-framework/modules/0L/Cases.move
```diff
@@ -29,10 +30,21 @@ address 0x1{
...
+            print(&height_end);
+            // this is a failure mode. Only usually seen in rescue missions, where epoch counters are reconfigured by writeset offline.
+            if (height_end < height_start) return VALIDATOR_DOUBLY_NOT_COMPLIANT;
+
             Roles::assert_diem_root(vm);
             // did the validator sign blocks above threshold?
+            print(&30031221);
+
             let signs = Stats::node_above_thresh(vm, node_addr, height_start, height_end);
+            print(&30031222);
```
Leibniz137, line 44: not a rustacean but i am wondering what these print statements do, they seem hardcoded. Is this a `print('got here1')` statement?

sm86 (approved):
fast track approve

PR #1073, opened 2022-04-18
DEPR [move] Ancestry of accounts AKA permission trees 

0o-de-lally:
Bring the onboarding history on-chain. 
With this information we can create a few more (imperfect) guarantees against sybil behavior. Possible uses are Validator set selection (i.e. Validator Readiness Program, knowing who has nodes ready to operate), or for Teams, for Carpe users qualifying to join a team.

- [x] Ancestry Module: init ancestry from onboarding in DiemAccount.
- [x] Ancestry: create admin migration script to use from when offline.
- [x]  Tests
- [x] implement ancestry in Vouch module: to identify which validators have other validators from different lineages vouching for them.


0o-de-lally:
Close, in favor of rebased version https://github.com/OLSF/libra/pull/1074.

PR #1083, opened 2022-04-21
Make Whole migration

0o-de-lally:
Refactors to do Make whole offline with a Writeset. Opportunity to refactor since we don't need to do it in-flight during an live epoch change.

- [x] writeset generator for make whole migration
- [x] change makewhole data structure, now per account
- [x] future proof for future incidents that require credits
- [x] keep same APIs
- [x] offline migration from fixture file make_whole.json
- [x] tests


language/diem-framework/modules/0L/MakeWhole.move
```diff
@@ -20,139 +22,102 @@ address 0x1 {
...
-            assert(*Vector::borrow<bool>(&payments.paid, i) == false, Errors::internal(EALREADY_PAID));
+            if (!exists<Balance>(addr)) return 0;
 
-            let amount = *Vector::borrow<u64>(&payments.amounts, i);
+            let b = borrow_global_mut<Balance>(addr);
+            let amount = 0;
+            let i = 0;
+            while (i < Vector::length(&b.credits)){
+              let cred = Vector::borrow_mut(&mut b.credits, i);
+              amount = amount + Diem::value<GAS>(&cred.coins);
```
jamesmeijers, line 61: You are summing the value in the loop, but also paying out each iteration of the loop. In the case where someone has multiple payments assigned (which I don't think should happen, but still), they will get a higher payout that they are supposed to (and the fund will be depleted before paying everyone out).

language/diem-framework/modules/0L/MakeWhole.move
```diff
@@ -20,139 +22,102 @@ address 0x1 {
         const EALREADY_PAID: u64 = 22017;
 
 
-        public fun make_whole_init(vm: &signer){
+        public fun vm_offer_credit(
```
jamesmeijers, line 25: Could I get an ELI5 for how this function is being run with both a vm signer and an account signer? Is it being called in a block prologue, or through some other mechanism?

language/diem-framework/modules/0L/MakeWhole.move
0o-de-lally: Nice catch!

0o-de-lally: This was spillover from a previous sketch I made. Fixed now.

codilion (approved):

PR #1086, opened 2022-04-22
Allows to send max allowed proofs per epoch

intuitiveminds:
## Motivation
Sending proofs stops 1 proof short of the maximum allowed number of proofs in an epoch due to an off-by-1 error in the check. This PR fixes that.


0o-de-lally (approved):
woot

PR #1088, opened 2022-04-22
Vouch tx

0o-de-lally:
Transaction tools for sending Vouch Tx

- [x] Move tx scripts for vouch
- [x] Rust txs subcommands for init_vouch and vouch_for


drawks:
This looks like it needs to rebased or something. PR looks too big for what it does and not enough information in the commits to understand the intent of the change.

0o-de-lally:
> This looks like it needs to rebased or something. PR looks too big for what it does and not enough information in the commits to understand the intent of the change.

@drawks This is an issue of the diem code, where a recompile of the Move code creates a lot of artifacts in the source. So to be sure, the only files changed are *.move and there are only a couple. So you can filter for those.

codilion (approved):

PR #1090, opened 2022-04-24
[move] Recovery mode for epoch boundary

0o-de-lally:
To better coordinate recovery efforts of a network, a recovery mode aims to also prevent front-running behavior from validators and miners.

- [x] Refactored DebugMode into separate module RecoveryMode.
- [x] Shuts off after epoch_ends period AND a minimum of 21 validators are present in the validator set.
- [x] Does not pay validators nor miners until the recover period is off.

Note: Does not prevent miners or validators from sending VDF proofs.


Leibniz137 (approved):

ashiiix (approved):

language/diem-framework/modules/0L/RecoveryMode.move
```diff
@@ -0,0 +1,90 @@
...
+
+    public fun maybe_remove_debug_at_epoch(vm: &signer) acquires RecoveryMode {
+      CoreAddresses::assert_vm(vm);
+      if (!exists<RecoveryMode>(CoreAddresses::VM_RESERVED_ADDRESS())) return;
+
+      let enough_vals = if (
+        Testnet::is_testnet() || 
+        StagingNet::is_staging_net()
+      ){ true }
+      else { (DiemSystem::validator_set_size() >= 21) };
```
jamesmeijers, line 48: Does this function get the current validator set size, or the number of eligible validators? if it's the former, then enough_vals will never be true unless the original fixed set is over 21, as I don't see a mechanism to add validators to the fixed set.

0o-de-lally: Good point.

My assumption is that there are 2 modes, and they are exclusive:
1. where you create a fixed set for debugging.
2. you are actually going to recover so the fixed set list is empty.

Originally I wrote #1 because it wasn't clear how we could jailbreak a stuck netowork. TL;DR we may just do without the fixed_set implementation. And REcoveryMode then only does the grace period before payments start again.

0o-de-lally: We could just have the fixed_case expire after # of epochs. So we keep this option for future cases

0o-de-lally: see now

jamesmeijers: Yes, this looks better

PR #1091, opened 2022-04-25
[txs] use managment costs instead of cheap costs for autopay batch TX

intuitiveminds:
## Motivation
Setting autopay via batch file leads to out-of-gas exceptions because cheap transaction cost settings are used instead of management transaction costs


0o-de-lally (approved):
LGTM

Issue #1107, opened 2022-05-11
[Bug]


0xzoz:
# 🐛 Bug

`is_slow_wallet` [function](https://github.com/OLSF/libra/blob/b7e8d062acb5e3cd167596e2c0db2ef9002e0322/ol/cli/src/node/query.rs#L359-L380) was introduced in #994 when unlocked balance was always 0 due to GAS not being unlocked by the protocol and flowing to slow wallets. The introduction of allowing the locked balance to be dripped broke this function.

## Expected Behavior

If a wallet is set as a slow wallet the function should return true. If it doesn't have a 0 balance this does not happen

## Steps To reproduce

run is_slow_wallet function from `ol/cli/src/node/query.rs`

0xzoz:
Merged into main

PR #1108, opened 2022-05-12
Fix query is_slow_wallet function due to breaking changes

0xzoz:
## Motivation

Fix for issue #1107

## Test Plan

You can run the following command

`cargo xtest --package ol`

## Related PRs

#994


0o-de-lally (approved):
LGTM

PR #1111, opened 2022-05-12
[tower] Allow tower to mine dynamic VDF parameters

0o-de-lally:
- [ ] change tower so that it can have local and online mode, where the next proof params are fetched from a query on chain (to be the default)
- [ ] add move resource type in rust to facilitate the query of VDFDifficulty
- [ ] move: refactor the VDF parameters so that they are on-chain

NOTE: Carpe and other tools need to implement API changes in their repos

### DIFF
--- a/language/diem-framework/modules/0L/TowerState.move
+++ b/language/diem-framework/modules/0L/TowerState.move
@@ -87,6 +87,22 @@ module TowerState {
         epochs_since_last_account_creation: u64
     }
 
+    struct VDFDifficulty has key {
+      difficulty: u64,
+      // Security parameter for VDF
+      security: u64,
+    }
+
+    // Create the difficulty struct
+    public fun init_difficulty(vm: &signer) {
+      CoreAddresses::assert_diem_root(vm);
+      move_to<VDFDifficulty>(vm, VDFDifficulty {
+        difficulty: 5000000,
+        security: 512,
+      });
+    }
+
+
     /// Create an empty list of miners 
     fun init_miner_list(vm: &signer) {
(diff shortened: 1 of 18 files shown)


0lBounties (approved):

0xzoz (approved):

Leibniz137 (approved):

language/diem-framework/modules/0L/TowerState.move
```diff
@@ -250,8 +296,8 @@ module TowerState {
       // Check vdf difficulty constant. Will be different in tests than in production.
       // Skip this check on local tests, we need tests to send differentdifficulties.
       if (!Testnet::is_testnet()){
-        assert(&proof.difficulty == &Globals::get_vdf_difficulty(), Errors::invalid_argument(130105));
-        assert(&proof.security == &Globals::get_vdf_security(), Errors::invalid_state(130106));
+        assert(&proof.difficulty == &Globals::get_vdf_difficulty_baseline(), Errors::invalid_argument(130105));
+        assert(&proof.security == &Globals::get_vdf_difficulty_baseline(), Errors::invalid_state(130106));
```
misko9, line 300: Should this be get_vdf_security_baseline() instead of difficulty?

0o-de-lally: Nice catch! Yes, but it was fixed in a subsequent PR.

The final is: https://github.com/OLSF/libra/blob/c6d6c65b8739cccb4b29152cc52ca9197d88fcb2/language/diem-framework/modules/0L/TowerState.move#L298

PR #1113, opened 2022-05-16
[mempool] Specify number of workers for mempool threads. Add prometheus metrics for networking.

0o-de-lally:
Part of an ongoing project of reducing the load which backlogs of VDF proofs produce on the network.

This mainly changes the concurrency behavior of the mempool, so that all mempool transactions use the same concurrency settings from Tokio Semaphore. Other operations were using this as a way of putting backpressure on transactions. For mempool tx verification this was not the case. And the hypothesis is that a large mempool queue of VDFs will cause resource exhaustion.

This PRS also adds more prometheus logs for state sync and mempool disconnection behavior.


intuitiveminds (approved):
LGTM

agouin (approved):

PR #1115, opened 2022-05-18
[diem-node] Patch Move error type handling by VM because of 0L error format changes

jamesmeijers:
This is in response to the issues faced by validators. I believe this is an important source of problems.

Recently, when my node fell out of the set, the last non-trivial error it had was: 
`ERROR language/diem-vm/src/errors.rs:95 [diem_vm] Unexpected prologue Move abort: 00000000000000000000000000000001::DiemAccount::1004 (Category: 236 Reason: 3) {"base_version":40969506,"name":"validation","txn_id":0}`

This led me to try find this error, which is PROLOGUE_EACCOUNT_DNE in DiemAccount. Notably, that error code is 1004, not 3, as would be suggested by the "Reason" given in the error. Additionally,  checking Errors.move tells be that "Category" 236 does not exist. 

erros.rs in the diem-vm handles errors assuming that they are in the Category, Reason format. However, 0L updated Errors.move to only give the Reason, to make our error output more straightforward, see [here](https://github.com/OLSF/libra/blob/837785dcfdaac2c6143a47775c6e5799e64cbcce/language/move-stdlib/modules/Errors.move#L20).  So, the diem-vm error handler is seeing an unexpected error, and so it is unable to match it (this matching is specific to the prologue and epilogue, as there are relatively few possible errors there). Because it can't match the error, it is throwing an error in the Rust code, despite the benign underlying issue (an account not existing but sending a transaction, in this case). 

I don't know what exactly the downstream effects of this error would be, or if it is the cause of the network issues, but there are a few things that lead me to think it might be the cause. 

- This would account for the intermittent nature of the failures. Only erroneous transactions would cause trigger this error, and so they should be rare and somewhat random. 
- The error would be more likely to be caused when more users are sending transactions/are online, which is what has been experienced
- Nodes should not propagate erroneous transactions, which would explain why only some validators are failing, again, seemingly at random 

I am currently running this on my node, and it appears stable thus far (but it's only been 20 minutes)

### DIFF
--- a/language/diem-vm/src/errors.rs
+++ b/language/diem-vm/src/errors.rs
@@ -60,48 +60,40 @@ pub fn convert_prologue_error(
             VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
         }
         VMStatus::MoveAbort(location, code) => {
-            let new_major_status = match error_split(code) {
-                (INVALID_STATE, EACCOUNT_FROZEN) => StatusCode::SENDING_ACCOUNT_FROZEN,
+            let new_major_status = match code {
+                (EACCOUNT_FROZEN) => StatusCode::SENDING_ACCOUNT_FROZEN,
                 // Invalid authentication key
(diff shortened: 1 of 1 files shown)


0o-de-lally (approved):
LGTM let's try this.

PR #1118, opened 2022-05-23
[move] Jail refactor. Resolve congestion of validator set, and resume set expansion

0o-de-lally:
TL;DR: 

The changes here fix the validator set selection algorithm and  introduce formal jailing, and un-jailing by a voucher.

A number of factors are preventing the validator set from expanding: a bug with queries, and sub-optimal sorting of validators. There's also a behavioral aspect, where the tools were allowing validators to un-jail themselves when they are actually unprepared to start.

Background:
Reducing the amount of seats available for validator set expansion causes congestion if there are several validators with high consensus power trying to join, but ultimately keep failing to stay in the set. The validator set cannot grow in this case. So the algorithm for picking validators based on Tower Height, needed a slight modification to privilege validators which have not been repeatedly failing to rejoin the validator set, and allow other prepared validators which may have a lower tower height from joining.

In Cosmos or other networks, it is usually the validator themselves which un-jail themselves. In 0L it has been that simply submitting enough Tower proofs will un-jail themselves. With this proposal, instead of automatically allowing to rejoin after tower proofs have been made, the Jail module benefits the network by allowing a validator to do necessary maintenance on the node before attempting to rejoin the validator set before they are ready, and having a Voucher check the work. 

The jail module also incorporates some non-monetary reputation staking features of Vouch. The Voucher has the incentive to check the work of the vouchee, because the Voucher's Jail state will track how many times their Vouchees have been jailed. This does not preclude a monetary penalty from also being implemented at a future date.

Changes:
1. Validators who get an is_jailed mark will not be considered for the next epoch, this was previously developed in the code since genesis but never activated.
2. To become un-jailed any of the Vouchers for a validator can execute an "un-jail" transaction for that validator.
3. Track how many attempts the node has made to join the set, but failed to stay in: consecutive_fail_to_join.
4. First the prospective validators by consecutive_fail_to_join, and then sub-sort by consensus power. Validators with consecutive_fail_to_join of 0, with highest tower height always have preference. A Validator with low tower height and consecutive_fail_to_join of 0, has preference over a high tower height and a consecutive_fail_to_join of 1.
5. Also track the lifetime jailing of the validator. Only for informational purposes.
6. Track the lifetime jailing of the Vouchees of the validator, recursively. Also only for informational purposes.

Fixes:
1. Algorithm was starting from a filtered list by Node Weight, when it should have included all prospective validators. Some validators with smaller towers, despite being ready were not able to join, since they were not included in top of the funnel.

Examples:
If Alice fails to validate at the threshold necessary (is either a Case 1, or Case 2), then her Jail state gets updated, with `is_jailed=true`. This means she cannot re-enter the validator set, until someone that vouched for Alice (Bob) sends an un_jail transaction. Bob has interest in seeing Alice rejoin and be performing since Bob's Jail state also gets updated based on Alice's performance.
On Bob's Jail state there is a `lifetime_vouchees_jailed`, every time Alice (because Bob has vouched for her) gets marked for jail, this number gets incremented on Bob's state too. Different than previously stated, it is NOT recursive to vouchers of vouchers.


language/diem-framework/modules/0L/Jail.move
```diff
@@ -0,0 +1,141 @@
...
+      // Bubble sort algorithm
+      let length = Vector::length(&vec_address);
+
+      let i = 0;
+      while (i < length){
+        let j = 0;
+        while(j < length-i-1){
+
+          let value_j = get_failure_to_join(*Vector::borrow(&vec_address, j));
+          let value_jp1 = get_failure_to_join(*Vector::borrow(&vec_address, j));
```
ashiiix, line 112: is there an issue on the indexes?

language/diem-framework/modules/0L/Jail.move
```diff
@@ -0,0 +1,141 @@
...
+    public fun vouch_unjail(sender: &signer, addr: address) acquires Jail {
+      // only a validator can un-jail themselves.
+      let voucher = Signer::address_of(sender);
+
+      let buddies = Vouch::buddies_in_set(addr);
+      let (is_found, _idx) = Vector::index_of(&buddies, &voucher);
+      assert(is_found, 100103);
+
+      // check the node has been mining before unjailing.
+      assert(TowerState::node_above_thresh(addr), 100104);
```
ashiiix, line 89: why is the mining check there? It will be done at epoch boundary right? One voucher might want to unjail while the validator has not submitted 8 proofs yet.

ashiiix (dismissed):

ashiiix:
In the PR description it says the jail count is recursive. I did not see it in the code, is that correct?
I have mixed opinion about the recursivity of the punishment as you may encounter cycles (infinite loops) and/or end up incrementing the counter on every validators, rendering this criterion meaningless.

language/diem-framework/modules/0L/ValidatorUniverse.move
```diff
@@ -22,10 +22,22 @@ address 0x1 {
         validators: vector<address>
     }
 
+    // deprecated
     struct JailedBit has key {
         is_jailed: bool
     }
 
+    struct Jail has key {
```
jamesmeijers, line 30: Why is this struct repeated here?

miskoio:
> In the PR description it says the jail count is recursive. I did not see it in the code, is that correct? I have mixed opinion about the recursivity of the punishment as you may encounter cycles (infinite loops) and/or end up incrementing the counter on every validators, rendering this criterion meaningless.
I agree, it seems the lifetime_vouchees_jailed value would skew towards active genesis validators having an increased amount since they are the root of the ancestry, especially if they were part of the April rescue party. Maybe a ding from the single voucher for unjailing, but I also don't mind not seeing lifetime_vouchees_jailed increase in the code. I'd prefer to see a jailing negate any potential tower increase, i.e NodeWeight = tower_height - (lifetime_jailed * 72).

intuitiveminds (approved):

language/diem-framework/modules/0L/Jail.move
0o-de-lally: Yes! nice catch. Fixed

language/diem-framework/modules/0L/Jail.move
0o-de-lally: Yeah I suppose that's a reasonable point. But we don't have this check actually at the epoch boundary, so we would need it here.

0o-de-lally:
> In the PR description it says the jail count is recursive. I did not see it in the code, is that correct? I have mixed opinion about the recursivity of the punishment as you may encounter cycles (infinite loops) and/or end up incrementing the counter on every validators, rendering this criterion meaningless.

Yes I think the recursive idea is not truly feasible unless you would build a PageRank algorithm. I think we can just Keep it at the direct vouchee, or maybe grandchild vouchee as well. I'll propose code for this.

language/diem-framework/modules/0L/ValidatorUniverse.move
0o-de-lally: fixed

bogdan-alexandrescu:
Given that lifetime_vouchees_jailed is tracked for informational purposes only, why will this be seen as a penalty for vouchers vouching for unstable validators? Are we planning to use the value of lifetime_vouchees_jailed later on to enact penalties?

0o-de-lally:
> Given that lifetime_vouchees_jailed is tracked for informational purposes only, why will this be seen as a penalty for vouchers vouching for unstable validators? Are we planning to use the value of lifetime_vouchees_jailed later on to enact penalties?

There's no plan. Though, the goal is to begin to track more performance data on-chain about the validators. It's very hard to get any analytics currently.

o0l0ogh:
I appreciate the intent behind this and feel it can be effective at this time. The aspect of consecutive_fail_to_join metric that doesn't quite sit well with me is that it could penalize a validator who truly believes they have everything figured out for *trying* to get into the set.

Speaking from my own experience, when I was unable to vote and didn't have a clear path to resolution, what I wanted was a mechanism to essentially flag myself as not being ready to enter the set. I understand this only applies to validators who are actively trying to get back into the set. It wouldn't apply to validators who are inactive, lost interest, etc.

However, without a test net to deploy potential fixes, re-entering the set is risky. For example, if a validator thought they had all the issues fixed, then tried to re-enter the set, only to discover the issues weren't fixed, they'd essentially get penalized for trying.

So maybe implement a "grace period", e.g. the first 2-3 efforts aren't counted, however those after that are. This could be an effective way to create some flexibility, while still filtering out uninterested/inactive validators.

And it feels hopeful to see 0L pushing the envelope on validator set admission/inclusion here!

0o-de-lally:
> I appreciate the intent behind this and feel it can be effective at this time. The aspect of consecutive_fail_to_join metric that doesn't quite sit well with me is that it could penalize a validator who truly believes they have everything figured out for _trying_ to get into the set.

This is a thoughtful response.

I would say that that consecutive_fail_to_join is not used as a penalty. There's no connection to slashing. It's just needed for raking which validators should get priority. It's resolving this issue of only using tower height (consensus power) for ordering entry into the validator set, which is currently preventing qualified validators from entering the set.

WRT flagging yourself to not be included. I think that's certainly something that can be added. The current way to do that, is just not send `tower` proofs.

jamesmeijers (approved):

Issue #65, opened 2020-06-22
0LIP-0065: Voting power needs to be updated with Reconfiguration


0o-de-lally:
In Libra system the voting power is hard coded to 1.
see here: 
https://github.com/OLSF/libra/blob/b8d061a4d8f9c65be2fc80cc3131deff3f535d90/language/stdlib/modules/LibraSystem.move#L95

In 0L reconfiguration sets the weights based on the proof-of-weight proposed in NodeWeight.move.

PR #1121, opened 2022-05-25
[txs] Require users to enter account address if they changed authkey

ea-open-source:
## Motivation
Users can change their account's authkey by using the command `txs authkey --new_key $NEW_AUTH_KEY`, however the current toolchains would not work with the new authkey. Reason being CLI requires users to enter their mnemonic and then derives authkey, keypair,  and account address from said mnemonics. But the new set of authkey/new mnomenic isn't linked with the old account address, so there is no way to get the correct account address.

**This PR requires the user to provide their mnemonic alongside with an optional old account address if they changed their account's address**

## Test Plan

None

## Related PRs

[Key rotation #65](https://github.com/OLSF/carpe/pull/65)

### DIFF
--- a/ol/txs/src/tx_params.rs
+++ b/ol/txs/src/tx_params.rs
@@ -1,28 +1,30 @@
 //! Txs App submit_tx module
 #![forbid(unsafe_code)]
-use crate::config::AppCfg;
+
+use std::io;
+use std::path::PathBuf;
+
 use anyhow::Error;
+use reqwest::Url;
 
 use diem_crypto::{
     ed25519::{Ed25519PrivateKey, Ed25519PublicKey},
     test_utils::KeyPair,
 };
 use diem_global_constants::OPERATOR_KEY;
-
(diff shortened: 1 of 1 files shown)


0o-de-lally:
Hello thanks for this PR. 
I don't think we want to be asking for user input in this module. For library type modules which are imported elsewhere (e.g. GUI) there will be no way to input data. Also I think we should follow Tell Don't Ask pattern here `https://martinfowler.com/bliki/TellDontAsk.html`.

There are two ways we should retrieve the user address:
1. From command line flags for each CLI app such as `--account`
2. Load from AppConfig using the 0L.toml

Also for future reference there is a rust library called Dialoguer.

0o-de-lally:
This looks correct thank you. Can you rebase this against the V6 branch?

0o-de-lally:
Admins, I think this can be merged. @sirouk @hemulin

sirouk:
LGTM

PR #1122, opened 2022-05-31
[tools] ol health will print local db version

0o-de-lally:
`ol health` and `ol start` to print local version.


0o-de-lally:
included in https://github.com/OLSF/libra/pull/1093

PR #1129, opened 2022-06-07
📝 Log rotation 

ownyrd:
## Motivation

The diem node + tower apps produce a lot of logs, so memory can fill up quickly without log rotation. I`ve added multilog as an example how log rotation can be achieved.


0o-de-lally:
very cool, thanks for this!

0o-de-lally (approved):
LGTM!

PR #1131, opened 2022-06-08
[tools] Explorer and Web Monitor are excluding good validators which may have autopay disabled

0o-de-lally:
This is an api bug, and not a system contracts bug. The list of validators is being incorrectly filtered when an autopay struct is not found.


agouin (approved):
works great!

PR #1133, opened 2022-06-09
[tower] Add garbage collecting functions to remove known invalid proofs.

0o-de-lally:
- [x] check for type of tx commit failure, and start GC
- [x] function to check for discontinuous proofs
- [x] archive files in separate folder


0o-de-lally:
merged into https://github.com/OLSF/libra/pull/1111

PR #1134, opened 2022-06-13
[web-monitor] Add burn prefs to API

0o-de-lally:
- [ ] API is missing a view for the validators to check their burn settings


agouin (approved):
Maybe we should .gitignore and remove the web-monitor build files? Otherwise looks good!

PR #1145, opened 2022-07-27
#1143 Lesson 3 Homework - Replace an unwrap() with match select statement

The1OneID:

Issue #1144, opened 2022-07-27
[Chore] Refactor `.unwrap()` with Match in Codebase


0xzoz:
# 👷🏾 Chore

> This should be done using the `v6` branch as the main has been frozen while we wait to upgrade the network.
> GRADUATION DEADLINE: To graduate from the Tool Scrubbers Guild's first course. You must submit [10 `.unwrap()` refactors](https://github.com/OLSF/libra/issues/1144) and [10 `.clone()` refactors](https://github.com/OLSF/libra/issues/1143). This should be submitted by 05/10/22.

## What is the issue

During the development of the initial implementation of 0L. Work was done quickly to meet deadlines and because of this `.unwrap()` was used in some instances to speed up development. These need to be cleaned up. 

## What needs to be done?

- Go through the `/ol` folder in the root and remove instances of `.unwrap()` by refactoring the code and adding a match statement. This may include refactoring of related files that the code is coupled to. 

## How to go about this?

If you take a section of this please open up a task in the tool-scrubbers-repo project board with the file or folder you are working on and comment below. This issue will add a checkbox here to keep track of what has been done.

- [X] `/ol/cli` - @0xzoz 
- [ ] `/ol/devnet`
- [ ] `/ol/documentation`
- [ ] `/ol/fixtures`
- [ ] `/ol/genesis-tools`
- [ ] `/ol/integration-tests`
- [ ] `/ol/keys`
- [X] `/ol/onboard` - @The1OneID 
- [ ] `/ol/tower`
- [X] `/ol/txs` - @nimadebi 
- [X] `/ol/types` - @MichaelZelbel
- [X] `/ol/util` - @mortonbits 
- [ ] `/ol/verifiable_delay`

The1OneID:
https://github.com/OLSF/libra/pull/1145

0xzoz:
I'll take care of the cli folder

The1OneID:
I'm working in the onboard folder

The1OneID:
#1151

MichaelZelbel:
I can try working on the types folder

nimadebi:
I'll work on txs folder

0xzoz:
@nimadebi @mortonbits @MichaelZelbel  @nControl88 @simyounghyun @The1OneID These tasks have been broken out into sub folders as separate issues to be more specific. Each folder now has a bounty attached to it for 2000 GAS. I will @ the owners in each of the issues and then close this

PR #1146, opened 2022-07-28
[Chore] Cleanup Unwrap Instances CLI Folder

0xzoz:
## Motivation

- Address issue #1144 and contribute to the tool scrubbers guild.
- This PR covers all folders within the `/ol/cli` folder 

## Test Plan

- Run `cargo build`
- Run `rustfmt`

## Related PRs

#1145 Covers one file

### DIFF
--- a/ol/cli/src/mgmt/management.rs
+++ b/ol/cli/src/mgmt/management.rs
@@ -51,7 +51,13 @@ fn spawn_process(
 ) -> std::io::Result<std::process::Child> {
     // Create log file, and pipe stdout/err
     let outputs = create_log_file(log_file);
-    let errors = outputs.try_clone().unwrap();
+    let errors = match outputs.try_clone() {
+        Ok(errors) => errors,
+        Err(e) => {
+            eprintln!("Error: {:?}", e);
+            exit(1);
+        }
+    };
 
     Command::new(binary)
         .args(args)
@@ -75,7 +81,13 @@ impl Node {
 
         // Start as validator or fullnode
         let conf = app_config();
-        let node_home = conf.workspace.node_home.to_str().unwrap();
+        let node_home = match conf.workspace.node_home.to_str() {
+            Some(s) => s,
+            None => {
+                eprintln!("Error: node_home is not set");
+                exit(1);
+            }
+        };
         let config_file_name = format!("{}validator.node.yaml", node_home);
 
         let child = if *IS_PROD {
@@ -90,9 +102,21 @@ impl Node {
                 "failed to run 'diem-node', is it installed?",
             )
         } else {
-            let project_root = self.app_conf.workspace.source_path.clone().unwrap();
+            let project_root = match self.app_conf.workspace.source_path.clone() {
+                Some(s) => s,
+                None => {
+                    eprintln!("Error: source_path is not set");
+                    exit(1);
+                }
+            };
             let debug_bin = project_root.join(format!("target/debug/{}", NODE));
-            let bin_str = debug_bin.to_str().unwrap();
+            let bin_str = match debug_bin.to_str() {
+                Some(s) => s,
+                None => {
+                    eprintln!("Error: target/debug/{} is not set", NODE);
+                    exit(1);
+                }
+            };
             let args = vec!["--config", &config_file_name];
             if verbose {
                 println!("Starting 'diem-node' with args: {:?}", args.join(" "));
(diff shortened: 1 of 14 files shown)


0o-de-lally:
wow wow wow @0xzoz lot's of good cleanups here!

ol/cli/src/mgmt/restore.rs
```diff
@@ -368,7 +463,13 @@ pub fn restore_epoch(db_path: &PathBuf, restore_path: &str, verbose: bool) -> Re
         .arg(db_path)
         .arg("epoch-ending")
         .arg("--epoch-ending-manifest")
-        .arg(manifest_path.to_str().unwrap())
+        .arg(match manifest_path.to_str() {
+            Some(path) => path,
+            None => {
+                eprintln!("Error: restore path is not a valid path");
```
simsekgokhan, line 469: manifest path instead of restore path?

ol/cli/src/mgmt/restore.rs
```diff
@@ -413,7 +514,13 @@ pub fn restore_transaction(
     if !manifest_path.exists() {
         println!(
             "ERROR: cannot find manifest path: {}",
-            manifest_path.to_str().unwrap()
+            match manifest_path.to_str() {
+                Some(path) => path,
+                None => {
+                    eprintln!("Error: restore path is not a valid path");
```
simsekgokhan, line 520: Same here and the next change

ol/cli/src/node/query.rs
```diff
@@ -667,11 +826,23 @@ fn test_is_community_wallet_should_return_false_if_missing_unfreeze_votes() {
 fn test_is_community_wallet_should_return_false_if_missing_consecutive_rejections() {
     let value = vec![
         (
-            Identifier::new("is_frozen").unwrap(),
+            match Identifier::new("is_frozen") {
```
simsekgokhan, line 829: How about using one line expect here instead of match? I believe, it should do the exact same as the following 5 lines.
`Identifier::new("is_frozen").expect("")`
https://doc.rust-lang.org/std/result/enum.Result.html#method.expect

You can also use in other places:
`Identifier::new("is_frozen").expect("my error msg")`

ol/cli/src/mgmt/management.rs
```diff
@@ -90,9 +102,21 @@ impl Node {
...
+                    eprintln!("Error: source_path is not set");
+                    exit(1);
+                }
+            };
             let debug_bin = project_root.join(format!("target/debug/{}", NODE));
-            let bin_str = debug_bin.to_str().unwrap();
+            let bin_str = match debug_bin.to_str() {
+                Some(s) => s,
+                None => {
+                    eprintln!("Error: target/debug/{} is not set", NODE);
```
simsekgokhan, line 116: `debug_bin` not set?

ol/cli/src/mgmt/management.rs
```diff
@@ -196,7 +232,13 @@ impl Node {
             };
 
             let debug_bin = project_root.join("target/debug/ol");
-            let bin_str = debug_bin.to_str().unwrap();
+            let bin_str = match debug_bin.to_str() {
+                Some(s) => s,
+                None => {
+                    println!("ERROR: can't start web-monitor in dev mode. It doesn't seem like you have workspace.source_path set in 0L.toml. Exiting.");
```
simsekgokhan, line 238: debug_bin not set?

ol/cli/src/mgmt/management.rs
```diff
@@ -238,9 +280,21 @@ impl Node {
                 "failed to run 'ol', is it installed?",
             )
         } else {
-            let project_root = self.app_conf.workspace.source_path.clone().unwrap();
+            let project_root = match self.app_conf.workspace.source_path.clone() {
+                Some(p) => p,
+                None => {
+                    println!("ERROR: can't start pilot in dev mode. It doesn't seem like you have workspace.source_path set in 0L.toml. Exiting.");
```
simsekgokhan, line 286: Shouldn't it be just "ERROR: It doesn't seem like you have workspace.source_path set in 0L.toml. Exiting." ?
The exit is not due to dev mode.

ol/cli/src/mgmt/management.rs
```diff
@@ -238,9 +280,21 @@ impl Node {
...
+                    exit(1);
+                }
+            };
             let debug_bin = project_root.join("target/debug/ol");
-            let bin_str = debug_bin.to_str().unwrap();
+            let bin_str = match debug_bin.to_str() {
+                Some(s) => s,
+                None => {
+                    println!("ERROR: can't start pilot in dev mode. It doesn't seem like you have workspace.source_path set in 0L.toml. Exiting.");
```
simsekgokhan, line 294: "Error: debug_bin not set" ?

ol/cli/src/mgmt/restore.rs
```diff
@@ -285,7 +368,13 @@ impl Backup {
 
         println!(
             "fullnode yaml created, file saved to: {:?}",
-            yaml_path.to_str().unwrap()
+            match yaml_path.to_str() {
+                Some(path) => path,
+                None => {
+                    eprintln!("Error: restore path is not a valid path");
```
simsekgokhan, line 374: "yaml_path" ?

ol/cli/src/node/query.rs
simsekgokhan: Seems to me, there are a few places that you cannot use expect - for all changes in this PR. You should be able to use expect instead of match for 90% of the `match`es in these code changes.

0xzoz:
Thanks for the review @simsekgokhan appreciate you taking the time! Still very new to this and using refactoring as a way to learn. I will go through and address these issues, it makes a lot more sense with these comments.

ol/cli/src/node/query.rs
0xzoz: Fixed a lot of these. Thanks for the explanation.

simsekgokhan:
@0xzoz Sure np. Thanks for the PR. 
Error handling is not a simple topic in Rust. This chapter is helpful: https://doc.rust-lang.org/book/ch09-00-error-handling.html. But I would read this topic in another detailed book as well.

PR #1149, opened 2022-08-07
Osx makefile support

Leibniz137:
## Motivation
Very small change, just so that we can build on os x easily, which is useful for doing some basic commands (eg `txs vouch ...`)

OS X doesn't have the `sha256sum` command installed by default, but it does have `shasum -a 265`

## Test Plan
On os x, ran `make bins`

```
...
bash -c "sha256sum language/diem-framework/staged/stdlib.mv || shasum -a 256 language/diem-framework/staged/stdlib.mv"
bash: sha256sum: command not found
b4403226e970f1e81609157234421f6fb0557ef09d69b70df227bbf385211897  language/diem-framework/staged/stdlib.mv
...
```


0o-de-lally (approved):
LGTM!

PR #1151, opened 2022-08-08
Improve error handling

The1OneID:
## Motivation

For our case it is better to print a custom message and exit rather than panic.

## Test Plan

cargo test

## Related PRs

### DIFF
--- a/ol/cli/src/cache.rs
+++ b/ol/cli/src/cache.rs
@@ -11,6 +11,7 @@ use std::{
     fs::{self, File},
     io::Write,
     path::PathBuf,
+    process::exit,
 };
 
 /// caching database name, to be appended to node_home
@@ -46,10 +47,20 @@ impl Vitals {
     /// reach the json cache
     pub fn read_json(node_home: &PathBuf) -> Vitals {
         let cache_path = get_cache_path(node_home);
-        let file = fs::File::open(cache_path).expect("file should open read only");
-        let deser: Vitals = serde_json::from_reader(file).expect("file should be proper JSON");
-
-        deser
+        let file = match fs::File::open(cache_path) {
+            Ok(f) => f,
+            Err(e) => {
+                println!("Could not open file: \nError {}", e);
+                exit(1)
+            },
+        };
+        match serde_json::from_reader(file) {
+            Ok(value) => value,
+            Err(e) => {
+                println!("Could not read json file: \nError {}", e);
+                exit(1)
+            }
+        }
     }
 
     /// write json cache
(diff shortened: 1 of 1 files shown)


The1OneID:
#1151

ol/cli/src/cache.rs
```diff
@@ -46,10 +47,20 @@ impl Vitals {
...
         let cache_path = get_cache_path(node_home);
-        let file = fs::File::open(cache_path).expect("file should open read only");
-        let deser: Vitals = serde_json::from_reader(file).expect("file should be proper JSON");
-
-        deser
+        let file = match fs::File::open(cache_path) {
+            Ok(f) => f,
+            Err(e) => {
+                println!("Could not open file: \nError {}", e);
+                exit(1)
```
0o-de-lally, line 54: Ok this is a great start. In this case we actually don't want the program to always exit. So let's change the return of this function into a Result<> and refactor the callers of the function.

0o-de-lally (changes requested):

ol/cli/src/cache.rs
The1OneID: So are we going to panic here and match in the callers?

The1OneID: The two introduced unwraps are just for testing. There's something breaking the test

ol/cli/src/server.rs
```diff
@@ -31,7 +31,7 @@ pub async fn start_server(mut node: Node, _run_checks: bool) {
         let interval = interval(Duration::from_secs(10));
         let stream = IntervalStream::new(interval);
         let event_stream = stream.map(move |_| {
-            let vitals = Vitals::read_json(&path);
+            let vitals = Vitals::read_json(&path).unwrap();
```
0o-de-lally, line 34: Instead of unwrap() here you want to use `?` like this Vitals::read_json(&path)?

The1OneID: That part is actually outdated, currently exiting. So you think is better to propagate it?

ol/cli/src/cache.rs
```diff
@@ -63,8 +62,8 @@ impl Vitals {
 
         // after writing temporary file renames and overwrite to cache file
         let cache_path = get_cache_path(node_home);
-        rename(temp_path, cache_path).expect("temporary cache file should be renamed");
-        Ok(())
+        let result = rename(temp_path, cache_path)?;
```
0o-de-lally, line 65: You can simply end the block of code with 

`rename(temp_path, cache_path)`

No need for the assignment and then the Ok()

ol/cli/src/server.rs
```diff
@@ -31,7 +31,13 @@ pub async fn start_server(mut node: Node, _run_checks: bool) {
         let interval = interval(Duration::from_secs(10));
         let stream = IntervalStream::new(interval);
         let event_stream = stream.map(move |_| {
-            let vitals = Vitals::read_json(&path);
+            let vitals = match Vitals::read_json(&path) {
+                Ok(v) => v,
+                Err(e) => {
+                    println!("Error reading json cache: \nError {}", e);
+                    exit(1)
```
0o-de-lally, line 38: I don't think we want to exit here. This would cause the web-monitor server to crash. Let's just return an error.

0xzoz (approved):

0xzoz (approved):

0xzoz:
@0o-de-lally need you to approve your requested changes before merging

PR #1152, opened 2022-08-10
[move] reduce thresholds on validators

0o-de-lally:
The validator set has struggled to grow because of some technical issues with state-sync and backup and restore. Though the thresholds, which exist to create a higher quality group of validators, are causing some confusion on debugging.

These thresholds should increase over time, as the tooling becomes more straightforward.
This PR proposes that the Vouch threshold be reduced from 4 to 2 vouches for validator, and the epoch block signing threshold be reduced from 5% to 3%.

These thresholds should increase again once a certain metric is hit in the network, e.g. 80+ validators successfully performing.

### DIFF
--- a/language/diem-framework/modules/0L/Globals.move
+++ b/language/diem-framework/modules/0L/Globals.move
@@ -118,7 +118,7 @@ module Globals {
           epoch_slow_wallet_unlock: 10,
           min_blocks_per_epoch: 0,
           vouch_threshold: 0,
-          signing_threshold_pct: 0,
+          signing_threshold_pct: 3,
         }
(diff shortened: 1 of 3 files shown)


sirouk:
LGTM

0xzoz:
This makes sense and also should help decrease inflation for the time being while holding some restrictions to make sure validators are active

LGTM!

0xzoz (approved):

sirouk (approved):

misko9 (approved):

The1OneID (approved):

PR #1153, opened 2022-08-11
[node] Don't panic on failure to initialize safety rules.

0o-de-lally:
We remove the panic!() which would prevent the node from starting in validator mode if the safety rules cannot be bootstrapped.

This is desirable because sometimes the validator needs to sync from a cold start, but the restore may not be perfect.
The definitive solution is in providing node operators with a restore point which does not fail.

The operator needs to be aware that doing this will cause an issue if the backup/restore artifacts have incomplete state. The operator will now see an Error in the logs, but will not panic.

To be clear: THE VALIDATOR WILL NOT BE ABLE TO SIGN BLOCKS if the safety riles cannot be initialized, but this change will allow it to sync.

### DIFF
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -302,9 +302,14 @@ impl EpochManager {
 
         let mut safety_rules =
             MetricsSafetyRules::new(self.safety_rules_manager.client(), self.storage.clone());
+        //////// 0L ////////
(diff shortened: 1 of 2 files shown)


sirouk (approved):
LGTM, thanks!

0xzoz (approved):

sirouk (approved):

sm86 (approved):
LGTM

agouin (approved):

PR #1157, opened 2022-08-17
cleanup unwrap ol/txs/src

simyounghyun:
## Motivation

Addressed issue(https://github.com/OLSF/libra/issues/1144)

## Related PRs

Covers part of ol/txs/src/entrypoint.rs

### DIFF
--- a/ol/txs/src/entrypoint.rs
+++ b/ol/txs/src/entrypoint.rs
@@ -175,7 +175,10 @@ pub fn get_args() -> EntryPointTxsCmd {
 /// in case of swarm like "....../swarm_temp/0" for alice
 /// in case of swarm like "....../swarm_temp/1" for bob
 pub fn get_node_home() -> PathBuf {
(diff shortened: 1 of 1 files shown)


ol/txs/src/entrypoint.rs
```diff
@@ -175,17 +175,26 @@ pub fn get_args() -> EntryPointTxsCmd {
 /// in case of swarm like "....../swarm_temp/0" for alice
 /// in case of swarm like "....../swarm_temp/1" for bob
 pub fn get_node_home() -> PathBuf {
-    let mut config_path = dirs::home_dir().unwrap();
+    let mut config_path = match dirs::home_dir() {
+        Some(r) => println!("Some"),
```
0o-de-lally, line 179: Thanks for this.
Here you want to be able to assign the config_path to the restult of the `match`. As written this is not returning the file path.

Use `cargo check` to check the compilation of this, and you will see some hints on how to solve it.

PR #1164, opened 2022-09-15
Refactor unwrap() with Match

nControl88:
First attempt to refactor unwrap() with Match

## Related PRs
#1144

### DIFF
--- a/ol/tower/src/proof.rs
+++ b/ol/tower/src/proof.rs
@@ -9,6 +9,7 @@ use glob::glob;
 use ol::node::client;
 use ol_types::block::VDFProof;
 use ol_types::config::AppCfg;
+use std::process::exit;
 use std::{fs, io::Write, path::PathBuf, time::Instant};
 use txs::tx_params::TxParams;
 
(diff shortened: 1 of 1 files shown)


ol/tower/src/proof.rs
```diff
@@ -286,9 +288,13 @@ fn create_fixtures() {
...
-            .write(&toml.as_bytes())
-            .expect("Could not write toml");
+        match file {
+            Ok() => file.write(&toml.as_bytes()),
+            Err(msg) => {
+                println!("Could not write toml\nError {}", msg);
+                std::process::exit(1)
+            }
+        }
     }
```
simsekgokhan, line 298: `std::process::exit` is a low level fn, should be used in specific cases.
https://doc.rust-lang.org/std/process/fn.exit.html ->
`
Note that because this function never returns, and that it terminates the process, no destructors on the current stack or any other thread’s stack will be run .... 
`
This means memory may not be cleaned at exit, hence big gateway for security/hacking problems.

Suggested way is to use `expect()` which uses `panic` internally instead of `exit` which is also short for match. So the code below actually is equivalent to two matches. One for file creation and one for write.
``` 
        file.expect("failed to create file")
            .write(&toml.as_bytes())
            .expect("Could not write toml");
```

See this also:
https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html

simsekgokhan:
Thx a lot for the effort, looks good to me. And if the main purpose of this task was not to exit or panic, then we would use match (or ? operator) to return the error to caller (or handle the error) instead of expect/unwrap which stops the program in case of error. FYI.

PR #1175, opened 2022-10-04
Refactor clone usage in CLI node folder

0xzoz:
## Motivation

This PR is to refactor some `.clones` as part of the Tool Scrubbers Guild Tasks. It contains 15 instances of `.clone()` cleanup. Some may not be applicable but would be good material for the Guild meeting. See comments 

## Test Plan

`cargo test` 

## Related PRs

N/A

### DIFF
--- a/ol/cli/src/node/account.rs
+++ b/ol/cli/src/node/account.rs
@@ -218,8 +218,8 @@ impl Node {
             Ok(account_state) => {
                 let handles = account_state.get_diem_account_resource()?.map(|resource| {
                     (
-                        resource.sent_events().clone(),
-                        resource.received_events().clone(),
+                        resource.sent_events().to_owned(),
+                        resource.received_events().to_owned(),
                     )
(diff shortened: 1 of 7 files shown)


0xzoz:
First commit [Reformat .clone to .owned](https://github.com/OLSF/libra/pull/1175/commits/a617d28e8f95cb31bbdaf68ae2a74b1323fcf258) was done due to some research. I found that when trying to use a reference for these EventHandles I could due to not being able to borrow a mutable reference twice. This may be a good discussion point for the guild based around lifetimes. Some resources I found are [here](https://dailydevsblog.com/troubleshoot/resolved-why-does-rust-try-to-borrow-self-as-mutable-again-when-it-already-holds-a-mutable-borrow-140872/) and [here](https://tutorialmeta.com/question/cannot-borrow-self-as-mutable-more-than-once-at-a-time-in-combination-with-h)

PR #1179, opened 2022-10-13
Added a step in smoke tests phase to cleanup space on runner machine

hemulin:
The CI pipeline under the github actions was failing on the smoke tests due to running out of space on runner machine. It is a bit surprising given that the runner machine supposedly has 14GB of space available, but anyhow :shrug: 

[Here](https://github.com/actions/runner-images/issues/2840#issuecomment-790492173) I found proposed solution.

Added a `space cleaning` step before the runner starts the bootstrapping and tested locally using a local github actions tester called [Act](https://github.com/nektos/act).
(which had other issues but I think the relevant part was passing correctly).

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -4,6 +4,12 @@ jobs:
   smoke:
     runs-on: ubuntu-latest
     steps:
+      - name: Clean up space on runner machine
+        run: |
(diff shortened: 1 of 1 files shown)


simsekgokhan (commented):
Looks good to me. Thx for the effort.

hemulin:
@simsekgokhan @0o-de-lally 
Approved? Can I merge?

PR #1182, opened 2022-10-24
Change "Out of gas" to "Gas bid too low"

MichaelZelbel:
This error message refers to the gas tank of an individual transaction. This is misleading. The message is exposed to the end user, who typically understands that his account is out of gas coins." He might not get the idea that he has to update the gas bid in his configs or in his application. When 0L gets popular, this might be frustrating for thousands of users.

## Test Plan

Trivial, low risk change. The automatic tests suffice.

### DIFF
--- a/api/types/src/convert.rs
+++ b/api/types/src/convert.rs
@@ -466,7 +466,7 @@ impl<'a, R: MoveResolver + ?Sized> MoveConverter<'a, R> {
                 AbortLocation::Script => format!("Move abort: code {}", abort_code),
             },
             KeptVMStatus::Executed => "Executed successfully".to_owned(),
-            KeptVMStatus::OutOfGas => "Out of gas".to_owned(),
+            KeptVMStatus::OutOfGas => "Gas bid too low".to_owned(),
             KeptVMStatus::ExecutionFailure {
                 location,
                 function,


hemulin (approved):
LGTM

hemulin:
@MichaelZelbel please wait for @simsekgokhan review before merging

simsekgokhan:
We also add this `//////// 0L //////// ` for our 0L changes in Diem files. 
e.g. 
https://github.com/OLSF/libra/blob/6d242b0b718ec49c2f382b9d0704c1d3a88e6c76/api/src/lib.rs#L5

@0o-de-lally do you also approve this change?

simsekgokhan:
@abellinii or @sm86 Can we get your review please?

api/types/src/convert.rs
```diff
@@ -466,7 +466,7 @@ impl<'a, R: MoveResolver + ?Sized> MoveConverter<'a, R> {
                 AbortLocation::Script => format!("Move abort: code {}", abort_code),
             },
             KeptVMStatus::Executed => "Executed successfully".to_owned(),
-            KeptVMStatus::OutOfGas => "Out of gas".to_owned(),
+            KeptVMStatus::OutOfGas => "Gas bid too low".to_owned(),
```
simsekgokhan, line 469: `//////// 0L //////// ` should be added at the end of line

simsekgokhan, line 469: `//////// 0L //////// ` should be added at the end of line

simsekgokhan (changes requested):

0o-de-lally:
Agree I think this is useful. @MichaelZelbel Please rebase on V6 and then I think it's ready to merge.

api/types/src/convert.rs
0o-de-lally: @MichaelZelbel yes please do this

PR #1210, opened 2022-11-09
Fix vm-genesis fn recovery_owners_operator()

simsekgokhan:
## 
Fixing panic (Move error) caused by "Set validator operator for each validator owner" section of `fn recovery_owners_operator`

https://github.com/OLSF/libra/blob/e042010ac22caba6cc9fc70c745b17f1bf7d666a/language/tools/vm-genesis/src/lib.rs#L839

In short, this is the fix: 
https://github.com/OLSF/libra/pull/1210/files#diff-0d5e9369eb26d02a79b4c768f68da66d258aee03b4f19409b5f7b3dc26d440d6L809-R814

tested and verified by cmd: 
`$ cargo run -p ol-genesis-tools -- --fork --output-path /opt/genesis_from_snapshot.blob --snapshot-path /opt/state_ver*`

where `/opt/state_ver` having these files: 
https://github.com/OLSF/epoch-archive/tree/main/359/state_ver_76353076.a0ff

## Details: 
The panic 
```
VMError { 
    major_status: NUMBER_OF_SIGNER_ARGUMENTS_MISMATCH, 
    sub_status: None, message: Some("Expected 2 signer args got 1"), 
    location: Undefined, indices: [], offsets: [] 
}, 
```

is caused by our code
`encode_set_validator_operator_with_nonce_admin_script_function`
This tx is not being used anywhere in original Diem code. 

----

This is how Diem code set the validator operator for each validator owner: 
https://github.com/OLSF/libra/blob/e042010ac22caba6cc9fc70c745b17f1bf7d666a/language/tools/vm-genesis/src/lib.rs#L644
and this is the tx being used: 

`encode_set_validator_operator_script_function`
https://github.com/OLSF/libra/blob/a116e3e75fd4f455b1ab9f5935b9f188c0baafff/language/tools/vm-genesis/src/lib.rs#L578

### DIFF
--- a/language/tools/vm-genesis/src/lib.rs
+++ b/language/tools/vm-genesis/src/lib.rs
@@ -185,7 +185,7 @@ pub fn encode_genesis_change_set(
 //////// 0L ////////
 pub fn encode_recovery_genesis_changeset(
     val_assignments: &[ValRecover],
-    operator_registrations: &[OperRecover],
+    operator_recovers: &[OperRecover],
     val_set: &[AccountAddress],
     // stdlib_modules: &[Vec<u8>],
     // vm_publishing_option: VMPublishingOption,
@@ -236,7 +236,7 @@ pub fn encode_recovery_genesis_changeset(
         &mut session,
         &log_context,
         &val_assignments,
-        &operator_registrations,
+        &operator_recovers,
         &val_set,
     );
     //////// 0L ////////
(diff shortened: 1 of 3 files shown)


simsekgokhan:
The problem is seen on macOS, not reproducible on Ubuntu.

simsekgokhan:
Correction: Reproducible both on macOS and Ubuntu.

0o-de-lally:
@simsekgokhan Is this still an issue with the current V6 head? Just please confirm it's needed since we merged the https://github.com/0LNetworkCommunity/libra/pull/1212

simsekgokhan:
> @simsekgokhan Is this still an issue with the current V6 head? Just please confirm it's needed since we merged the #1212

Just tested both on macOS and Ubuntu, the issue is not reproducible anymore, closing. thx.

PR #1211, opened 2022-11-24
Remove slow wallet check when completing a community transfer

0xzoz:
## Motivation

At the moment, community wallets can send GAS only to slow wallets. It doesn’t make sense. 
Slow wallets were made to prevent validators and early contributors that were granted large amounts of GAS, from dumping on the market and move the price too much.

People that are receiving bounties, “employees“, etc. all of those, should have no limitations to spend their “well earned“ GAS.

This PR removes the checks on the payee acc being a slow wallet when making community transfers.

## Test Plan

Refer to comment below

## Related PRs

N/A

### DIFF
--- a/diem-move/diem-framework/core/transactional-tests/0L/diem_account/transfer_community_wallet_script.move
+++ b/diem-move/diem-framework/core/transactional-tests/0L/diem_account/transfer_community_wallet_script.move
@@ -2,20 +2,20 @@
 // Alice, Jim:     validators with 10M GAS
 // Bob, Carol: non-validators with  1M GAS
 
-// Bob, the slow wallet
+// Bob, the recipient wallet
 // Carol, the community wallet
 
-// Community wallets cannot use the slow wallet transfer scripts
 
 //# run --admin-script --signers DiemRoot Bob
 script {
   use DiemFramework::DiemAccount;
   use Std::Vector;
 
-  fun main(_dr: signer, bob: signer) {
+  fun main(_dr: signer, _bob: signer) {
     // Genesis creates 6 validators by default which are already slow wallets,
     let list = DiemAccount::get_slow_list();
-    assert!(Vector::length<address>(&list) == 7, 735701);
+    //verify bob is not in the slow list
+    assert!(Vector::length<address>(&list) == 6, 735701);
   }
 }
 
--- a/diem-move/diem-framework/core/transactional-tests/0L/diem_account/transfer_community_wallet_script.move
+++ b/diem-move/diem-framework/core/transactional-tests/0L/diem_account/transfer_community_wallet_script.move
@@ -14,8 +14,6 @@ script {
 
   fun main(_dr: signer, bob: signer) {
     // Genesis creates 6 validators by default which are already slow wallets,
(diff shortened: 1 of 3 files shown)


0xzoz:
@simsekgokhan what is the best way to test these changes?

diem-move/diem-framework/core/transactional-tests/0L/diem_account/transfer_community_wallet_script.move
```diff
@@ -14,8 +14,6 @@ script {
 
   fun main(_dr: signer, bob: signer) {
     // Genesis creates 6 validators by default which are already slow wallets,
-    // adding Bob
-    DiemAccount::set_slow(&bob);
```
hemulin, line 18: @0xzoz 
Bob is a validator. As such, he should be set to slow.
Why removing this here?

0xzoz: Its a test script. Bob is not a validator in this instance

https://github.com/OLSF/libra/blob/36439a7064b3f0bc3b5cad0641aac17e88884224/diem-move/diem-framework/core/transactional-tests/0L/diem_account/transfer_community_wallet_script.move#L3

simsekgokhan:
> @simsekgokhan what is the best way to test these changes?

This is the test cmd, I will check the code a bit later.

`NODE_ENV="test" cargo test -p diem-framework --test ol_transactional_tests transfer_community_wallet_script.`

simsekgokhan (changes requested):
LGTM. Thank you! 
You need to update the test to make it pass. Lines 10, 15, 18 in your version or better, just see/follow the error codes of failed test.

0xzoz:
> You need to update the test to make it pass. Lines 10, 15, 18 in your version or better, just see/follow the error codes of failed test.

Thanks for the insight into how to test. The changes have been made and it is now passing.

I would like to hear @0o-de-lally's opinion on the implications of these changes from an economical perspective. I see the reasoning behind this being beneficial to recipients but fear I may be missing context to the other side of the argument.

0o-de-lally:
So I would strongly advise against this. This breaks a lot of the game theory around community wallets, where it makes it trivial to steal from the "match burn" donation game. Also it will completely change the expectations around vesting and circulation, which leads to other second order consideration.

I think that the daily drip can change once there's a governance discussion about it. But I think that at the current 30K per month [edit], I don't really see a major problem.

That all said, in V6 the actual numbers may need to change to achieve the original objective: vesting for all, especially whales, but not more that 10 years for the largest account. I believe this written up in one of the canonical docs.

0xzoz:
Thanks for the context @0o-de-lally. I had a feeling there would be missing information and/or indirectly affected aspects of the design but had a very naive understanding and wasn't sure what they were.


What is the 30K daily unlock you are referring to? The original documentation stated 10k per slow wallets address but the commented code had 1k, which is what is currently in place.

0o-de-lally:
> What is the 30K daily unlock you are referring to? 
I meant monthly. Edited now.

PR #1218, opened 2022-12-30
recent transactions and timestamps

agouin:
## Motivation

Optimize explorer queries and pagination, specifically fetching most recent transactions for both chain and an individual account.

Adds `timestamp_usecs` to `TransactionView` and `EventView` so that timestamps can be rendered in explorer for transactions and events without extra effort on explorer side.

Adds `get_recent_transactions` and `get_recent_account_transactions` RPC methods to fetch most recent transactions for chain and account.

## Test Plan

Tested with explorer changes https://github.com/agouin/0l-explorer/pull/6


0o-de-lally:
This looks quite helpful. Can you try rebasing to see if the tests pass.

PR #1220, opened 2023-01-12
main: Added `nix` shell script to make building the binaries easy on…

corynthian:
… `NixOS`.

## Motivation

The `Makefile` is currently specific to `Ubuntu`. This change adds a `NixOS` script which pulls in all the deps required to build the codebase. 

## Test Plan

Run `nix-shell` in the root on `NixOS` then run:
```
cargo build -p diem-node -p tower -p backup-cli -p ol -p txs -p onboard --release
```

If running for the first time run `rustup update` prior to the `cargo build` command.

## Related PRs

None?

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,56 @@
+{ pkgs ? import <nixpkgs> {} }:
+let
+  my-python = pkgs.python3;
+  python-with-my-packages = my-python.withPackages (p: with p; [
+    pip
+    Fabric
+  ]);
+in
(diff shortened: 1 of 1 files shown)


corynthian:
Just tried it on a new host, for some reason had to run `rustup install 1.52.1` from the `rust-toolchain` after `rustup update`.

0o-de-lally:
Super helpful. If you don't need it in the root of the project,  I would move it to ol/util where we have some .sh helpers.

PR #1223, opened 2023-01-15
v6: Merkle tree hashing

corynthian:
## Motivation

Implements `Keccak` (sha3-256) based Merkle Tree hashing.

Note: Also includes fixes to `VoteTests`.

## Test Plan

`cargo test --test move_unit_test`

## Related PRs

#627

### DIFF
--- a/diem-move/diem-framework/DPN/sources/0L/MerkleHash.move
+++ b/diem-move/diem-framework/DPN/sources/0L/MerkleHash.move
@@ -1,0 +1,77 @@
+/// Implements a Merkle Tree Hash using SHA3-256.
+module DiemFramework::MerkleHash {
+    use Std::Vector;
+    use Std::Hash;
+
+    /// An element was added past the max supported size.
+    const ELENGTH: u64 = 0;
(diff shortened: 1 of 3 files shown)


hemulin:
@ashiiix note that this is addressing [a ticket](https://github.com/0LNetworkCommunity/libra/issues/627) you were involved in before

corynthian:
One issue with this PR is that it uses `sha3-256` from `tiny-keccak` rather than the Ethereum `Keccak256`. So its not usable for an ETH bridge.

0o-de-lally:
This is stellar Move code! Hat's off!

We have a user here @coin1111 that has implemented a early version of an eth bridge. For that he created a "native function" with keccak. Although, oddly I don't see it in V6, so I think it was ignored with the migration.

Anyhow. Perhaps you want to pull in the XHASH module from V5, see below. And secondly, you might want to review the bridge code. Although, where we are today, it seems more likely that axelar would be the first bridge on 0L.

https://github.com/0LNetworkCommunity/libra/blob/main/language/diem-framework/modules/0L/xhash.move

PR #1226, opened 2023-01-27
[Testing Infra] V6 genesis recovery state test suite

0o-de-lally:
Adding functional tests for `ol-genesis-tools` for:
- [x] parsing a db backup, 
- [x] exporting a JSON intermediary file, 
- [x] generating a blob, 
- [x] and finally starting a test node from that blob.
- [x] plus one e2e test which does the whole flow.
- [x] test a single validator import from json and check the genesis matches
- [x] * test a sample recovery file, and check the genesis matches

* the test infrastructure is done, but the test is failing.


0o-de-lally:
R4R @simsekgokhan @nelaturuk @sirouk @corynthian

simsekgokhan:
> R4R @simsekgokhan @nelaturuk @sirouk @corynthian

LGTM

ol/genesis-tools/src/fork_genesis.rs
```diff
@@ -83,19 +89,33 @@ pub fn get_baseline_genesis_change_set(
...
     let mut total_coin_value = 0u64;
+    let mut len = 0u64;
+    // iterate through all the accounts
+    // this includes the Validator accounts.
+    // at a minimum balances will need to be recovered
+    // and the total tokens calculated, and checked for matching sums.
+    // there may also be migrations on other account state that may need to be done.
     for l in &legacy_vec {
+        assert!(l.account.is_some());
+        println!("migrating: {} - {}", &l.account.unwrap(), &len);
```
corynthian: Maybe `diem_logger::info` or `debug` here?

ol/genesis-tools/src/lib.rs
```diff
@@ -9,9 +9,9 @@
 )]
 
 pub mod fetch_archive;
-pub mod fork_daemon;
+// pub mod fork_daemon;
```
corynthian: Maybe here add a comment as to why this is commented out (e.g. if the module should be kept) or if the intent is to leave a trace of changes, maybe use the same convention as in the rest of codebase (`// 0L --- deleted: pub mod fork_daemon due to being replaced by foo`).

corynthian (commented):
Minor nitpicks, other than these (maybe I missed something) but I am not seeing where the account state post initialization is compared with the state used for recovery (e.g. checking that balances are the same / state is the same as expected).

ol/genesis-tools/src/lib.rs
0o-de-lally: yes good point I'll document that. In this case it's a feature that won't ship, at least not in this version.

ol/genesis-tools/src/fork_genesis.rs
0o-de-lally: Yeah would be good to use a logger. Good catch. We've often run into issues with using diem_logger externally. Will test.

0o-de-lally:
> Minor nitpicks, other than these (maybe I missed something) but I am not seeing where the account state post initialization is compared with the state used for recovery (e.g. checking that balances are the same / state is the same as expected).

Good call. Yeah checking the state should be done (I have a `todo:` somewhere). For now I'll do a single query, since real tests of this need some better introspection tools.

The best testsuite to do that is Forge or the higher level Smoke Tests. Though I'm not entirely sure how to start a Forge instance with an arbitrary Genesis blob. Maybe a volunteer of @simsekgokhan can figure this out?

0o-de-lally:
@corynthian I've pushed the changes you asked for including an onchain verifcation.

@simsekgokhan Please confirm that running `diem-node --test` actually uses all the genesis data, or only the stdlib code. I think this may not be the best way to check onchain state. See my changes to `e2e_test.rs`.

I think Forge, or Smoke test may be the only way to introspect this correctly. But I think some refactoring is needed to  have it use a genesis.blob

0o-de-lally:
@sirouk @0xzoz @hemulin I think this is ready to merge. I depend on it for the next PR https://github.com/0LNetworkCommunity/libra/pull/1231

0xzoz (approved):

0o-de-lally:
@0xzoz can you merge?

0xzoz:
> @0xzoz can you merge?

I dont have write access. It would be good to get one more approval as well. My eyes are not as well trained as others🙂

PR #1231, opened 2023-02-07
[Genesis] V6 refactor account state migration to use VM

0o-de-lally:
The chain fork state migration pattern is currently hand crafting writesets. This can lead to bad state, and doesn't enforce any conditions. The best practice is to start an Executor instance, and call into certain functions, and then preserve the Writesets from those operations. Then we can add those Writesets to a clean genesis (the minimal bootstapping transactions e.g. validator set election).

- [ ] Create module to start a FakeExecutor instance that calls arbitrary functions, and returns writesets
- [ ] Merge a baseline genesis transactions (i.e. its writesets) to the migration writesets
- [ ] Replace previous migration scripts to new pattern
  - [ ] validator settings (perhaps it's ok since it uses genesis tx tools?)
  - [ ] account balance
  - [ ] TowerState
  - [ ] Autopay
  - [ ] Receipts
  - [ ] Community Wallets (note: split Donor Directed, and Community Wallet)
  - [ ] Ancestry
  - [ ] Fullnode Subsidy (rename to Oracle Subsidy)
  - [ ] Burn
  - [ ] Slow Wallet & unlocked balance (note: check unlocking logic)
 - [ ] New structs
   - [ ] Infra escrow pledges
   - [ ] Founder ID (bloom filter)
   - [ ] Coin Split
 - [ ] Not migrating:
   - [ ] Vouches
   - [ ] MakeWhole
- [ ] Make new tests infra confirm genesis for all items (from https://github.com/0LNetworkCommunity/libra/pull/1226)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6571,6 +6571,7 @@ dependencies = [
  "diem-wallet",
  "futures",
  "gumdrop 0.8.1",
+ "language-e2e-tests",
  "move-core-types",
  "ol",
  "ol-keys",
--- a/ol/genesis-tools/Cargo.toml
+++ b/ol/genesis-tools/Cargo.toml
@@ -25,6 +25,7 @@ gumdrop = "0.8.0"
 ol = { path = "../cli/"}
 ol-keys = { path = "../keys/", version = "0.1.0" }
 ol-types = { path = "../types/" }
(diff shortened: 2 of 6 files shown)


sirouk:
LGTM

PR #1237, opened 2023-02-12
[Genesis] Migration tools for genesis from legacy accounts

0o-de-lally:
This is the tool which allows for a "kitchen sink" network upgrade, that is, one that requires a halt and a new genesis from backed up data.

Genesis tools for migration
- [x] Refactor account recovery to use the vm session in vm-genesis/lib.rs (quite gross work)
- [x] Remove daemon, file fetcher, swarm, which are not useful in the case of a single V6 upgrade.
- [x] Add test scaffold, to create a default test node config, and replace with current genesis.

Tests:
- [x] e2e tests where a test node goes up
- [x] parsing snapshot tests
- [x] exporting json tests
- [x] creating genesis from json
- [x] test with production snapshots

A separate and future issue would be how to properly test this in Forge and Smoke tests. Hard-ish problem.


ol/genesis-tools/Cargo.toml
```diff
@@ -27,6 +27,27 @@ ol-keys = { path = "../keys/", version = "0.1.0" }
...
+#for testing only
+diem-json-rpc = { path = "../../json-rpc" }
+reqwest = { version = "0.11.2", features = ["blocking", "json"], default_features = false }
+# genersis blob reading
+diem-genesis-tool = { path = "../../config/management/genesis", version = "0.1.0" }
+diem-temppath = { path = "../../crates/diem-temppath", version = "0.1.0" }
+move-binary-format = { git = "https://github.com/OLSF/move-0L", rev = "94e1039c9fdf1472c2c7775de43135d28dafc955" }
+move-vm-runtime = { git = "https://github.com/OLSF/move-0L", rev = "94e1039c9fdf1472c2c7775de43135d28dafc955" }
+move-bytecode-utils = { git = "https://github.com/OLSF/move-0L", rev = "94e1039c9fdf1472c2c7775de43135d28dafc955" }
+move-vm-types = { git = "https://github.com/OLSF/move-0L", rev = "94e1039c9fdf1472c2c7775de43135d28dafc955" }
```
0xzoz, line 42: Should these point to the new org url?

0xzoz (approved):

ol/genesis-tools/Cargo.toml
0o-de-lally: yes thanks for pointing out. The URL does redirect though

sirouk:
LGTM 👍

PR #1254, opened 2023-04-14
Donor Directed Accounts Liquidation

0o-de-lally:
Donor directed wallet liquidation case was incomplete.

- [x] Donor directed wallets can be liquidated by donors with ParticipationVote like Veto of transactions.
- [x] In default case the funds return to the donor wallets
- [x] In the community wallet case, the funds go to TransactionFee so it follows the usual burn/recycle governance.
- [x] The VM needs to execute this on the epoch boundary
- [x] Tx tests for whole flow of liquidation


0xzoz:
Closing due to changes included #1261

PR #1264, opened 2023-08-30
[Tools] Recovery ancestry patch

0o-de-lally:
For the V5->V6->V7 migration we need to generate migration files from this code base. The `Ancestry` account metadata needs to be patched on a number of accounts.


sirouk (approved):
LGTM 
Nice cleanup on formatting too 👍

